PR #30, opened 2015-01-12
Add support for resolving base URLs in <base> tags with attributes before href.

ArturGaspar:
This adds support for finding the base URL when the <base> tag has some other attribute before the href one. According to http://www.w3.org/TR/html5/document-metadata.html#the-base-element and by my understanding http://www.w3.org/TR/html401/struct/links.html#h-12.4 too, the base element can have other attributes as well.

### DIFF
--- a/w3lib/html.py
+++ b/w3lib/html.py
@@ -13,7 +13,7 @@ from w3lib.url import safe_url_string
 
 _ent_re = re.compile(r'&(#?(x?))([^&;\s]+);')
 _tag_re = re.compile(r'<[a-zA-Z\/!].*?>', re.DOTALL)
-_baseurl_re = re.compile(six.u(r'<base\s+href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
+_baseurl_re = re.compile(six.u(r'<base[^>]+href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
 _meta_refresh_re = re.compile(six.u(r'<meta[^>]*http-equiv[^>]*refresh[^>]*content\s*=\s*(?P<quote>["\'])(?P<int>(\d*\.)?\d+)\s*;\s*url=(?P<url>.*?)(?P=quote)'), re.DOTALL | re.IGNORECASE)
 _cdata_re = re.compile(r'((?P<cdata_s><!\[CDATA\[)(?P<cdata_d>.*?)(?P<cdata_e>\]\]>))', re.DOTALL)
 
--- a/w3lib/html.py
+++ b/w3lib/html.py
@@ -13,8 +13,8 @@ from w3lib.url import safe_url_string
 
 _ent_re = re.compile(r'&(#?(x?))([^&;\s]+);')
 _tag_re = re.compile(r'<[a-zA-Z\/!].*?>', re.DOTALL)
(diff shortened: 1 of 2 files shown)


w3lib/html.py
```diff
@@ -13,7 +13,7 @@
 
 _ent_re = re.compile(r'&(#?(x?))([^&;\s]+);')
 _tag_re = re.compile(r'<[a-zA-Z\/!].*?>', re.DOTALL)
-_baseurl_re = re.compile(six.u(r'<base\s+href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
+_baseurl_re = re.compile(six.u(r'<base[^>]+href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
```
kmike: it seems that `<basefoo href="..">` is captured by this regex

kmike: as well as `<base foohref="..">`

kmike:
A good catch!

ArturGaspar:
A good catch indeed. But I based myself on the already existing `_meta_refresh_re` right below it. The same happens with `get_meta_refresh()`. Is it a bug?

    >>> import w3lib.html
    >>> w3lib.html.get_meta_refresh('<metafoohttp-equiv="refresh"content="1;url=http://example.com/">')
    (1.0, 'http://example.com/')
    >>> w3lib.html.get_meta_refresh('<metafoo http-equiv="refresh" content="1;url=http://example.com/">')
    (1.0, 'http://example.com/')

kmike:
"a good catch" was your catch about then `<base>` tag :)

Right, it looks like `<meta refresh` regexp has a similar problem.

ArturGaspar:
Ah, right, thank you. I found it while working with Scrapy to crawl a website.

ArturGaspar:
This should fix both.

w3lib/html.py
```diff
@@ -13,8 +13,8 @@
 
 _ent_re = re.compile(r'&(#?(x?))([^&;\s]+);')
 _tag_re = re.compile(r'<[a-zA-Z\/!].*?>', re.DOTALL)
-_baseurl_re = re.compile(six.u(r'<base\s+href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
-_meta_refresh_re = re.compile(six.u(r'<meta[^>]*http-equiv[^>]*refresh[^>]*content\s*=\s*(?P<quote>["\'])(?P<int>(\d*\.)?\d+)\s*;\s*url=(?P<url>.*?)(?P=quote)'), re.DOTALL | re.IGNORECASE)
+_baseurl_re = re.compile(six.u(r'<base\s[^>]*href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
+_meta_refresh_re = re.compile(six.u(r'<meta\s[^>]*http-equiv[^>]*refresh[^>]*content\s*=\s*(?P<quote>["\'])(?P<int>(\d*\.)?\d+)\s*;\s*url=(?P<url>.*?)(?P=quote)'), re.DOTALL | re.IGNORECASE)
```
kmike: it seems that _meta_refresh_re has other problems: 
* http-equivrefreshcontent is accepted;
* order of attributes is hardcoded

kmike: these problems are not related to this PR though

PR #33, opened 2015-04-28
Update meta_refresh regex

Djayb6:
I updated the meta_refresh regex to handle spaces, newlines and other characters supported by `\s` between `url=` and the URL

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -344,3 +344,12 @@ class GetMetaRefreshTest(unittest.TestCase):
             </html>"""
         self.assertEqual(get_meta_refresh(body, baseurl), (None, None))
 
+    def test_leading_newline_in_url(self):
+        baseurl = 'http://example.org'
+        body = """
+        <html>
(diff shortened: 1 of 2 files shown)


dangra:
Hi @Djayb6 , thanks for your fix. Please add test cases to back your changes and link to real world example in issue description. thanks

Djayb6:
Hello @dangra , here is the test case

dangra:
LGTM.  Please squash so I can merge. thanks.

Djayb6:
@dangra squashed !

dangra:
@kmike @Curita do you want to take a look before merging?

kmike:
Looks good!

Curita:
+1, but I think that regex should be extended with support of spaces after every token sometime.

kmike:
@Curita these regexes have other issues, see e.g. https://github.com/scrapy/w3lib/pull/30#discussion_r22810722

Curita:
Ohh, right, seems to require more thought into it then, thanks for pointing those comments!

PR #34, opened 2015-06-24
ENH: do not exclude tests from distribution

yarikoptic:
Those are handy things to be distributed so users (or packagers such as yours truly) could verify correct operation of the library on their system... if accepted here I would appreciate if similar change was applied to scrapy as well

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,3 @@
+# Include tests into distribution
+recursive-include tests *
+


kmike:
It will install 'tests' package to user packages, won't it?

yarikoptic:
d'oh -- you are totally right @kmike -- I rushed ;)  what about starting to use yet another obscure feature of python distribution tools? (pushed another solution involving MANIFEST.in)

yarikoptic:
in general, I prefer (and it seems to be generally more common) when tests come as a part of the Python module itself (e.g. here under w3lib/tests).  That allows people to test application "as installed".  I just didn't want to impose anything. But now that I have introduced MANIFEST.in ... ;-)

Digenis:
> in general, I prefer (and it seems to be generally more common) when tests come as a part of the Python module itself

Looks like the opposite happens here:
https://github.com/scrapy/w3lib/commit/ac08e09a72695da5a81d78b11ffc2e3a04f23182
https://github.com/scrapy/scrapy/commit/242c08553f5bed2aedf2ff962fc83f68e13f2259
btw pytest's documentation accepts both schemes as correct

dangra:
We tried the approach of shipping tests suite with the code, specially for Scrapy, but eventually problems arises like needing different dependencies to run tests than to run main app (i.e. mock, mitmproxy, py.test, ...).

I don't see distribution shipping tests suites of packages like apache, squid, bash, and so on, why scrapy related packages should be an exception?. I am aware distribution run test suites while building its packages, that's ok and can be done right but packaging test suites so they are available in final systems is a different and unlikely to happen topic.

To be honest the idea is noble but not practical.

yarikoptic:
I hear you @dangra !  My reply is just for completeness -- I would respect any decision you take... hopefully at least you would ship tests/ within source distribution.

Additional dependencies are indeed a hassle, and different projects deal with it in various ways, most common approach just skip specific tests (or entire test module) if necessary dependency is absent (e.g. in PyMVPA we can run tests reliably without any external dependency but numpy although we have probably around 10 optional dependencies).  Some (e.g. psychopy) also include complete py.test runner along to simplify running on some exotic systems without py.test being readily available.

As for comparison to other projects -- don't forget that we are quite special, since we are using Python, which come with builtin unittests module.  Relatively small feature but imho it is the one fostered culture of better testing (on average ;-) ) among its projects.  So, to my experience, majority of Python projects (numpy, scipy, pandas, sklearn, PyMVPA, even fail2ban now ;-), ...long list) ship with tests either under module/tests or even spread out per each submodule.  Many of them even provide "module.test()" call-in to trigger testing.
Complementary to this,  relying on tests being provided, such initiatives as DEP8 in the Debian world (http://dep.debian.net/deps/dep8) now blossoms with http://ci.debian.net/ which allows to identify causes for breakage while considering entire OS integration.
Hopefully this convinces you that it is actually quite practical.  I agree though that it does require a tiny bit more of work.

dangra:
I see your point, and aside of increase in tarball size I see no objections to ship _tests/_ dir in *source* .tar.gz distribution published to pypi. Right now for Scrapy we include _docs/_ and _extras/_ directories so nothing prevents us from including _tests/_ too. As long as binary distribution (wheel pkg) doesn't pack tests directory I am OK.

dangra:
LGTM.  @kmike feel free to merge.

yarikoptic:
Cool! Thanks!  Then for the next release I will enable package build time testing for Debian package.
Should I submit similar one against scrapy or ... ? ;)

dangra:
> Cool! Thanks! Then for the next release I will enable package build time testing for Debian package.

That's great. thanks!

> Should I submit similar one against scrapy or ... ? ;)

:+1: go for it.

PR #37, opened 2015-09-28
Remove uppercase tags

Youwotma:


w3lib/html.py
```diff
@@ -183,7 +183,11 @@ def remove_tags(text, which_ones=(), keep=(), encoding=None):
 
     assert not (which_ones and keep), 'which_ones and keep can not be given at the same time'
 
+    which_ones = set(tag.lower() for tag in which_ones)
+    keep = set(tag.lower() for tag in keep)
```
kmike: we can use `{tag.lower() for tag in which_ones}` because python 2.6 is not supported

dangra: +1 to new set syntax

PR #38, opened 2015-09-28
[MRG+1] Redirect inside noscript

Youwotma:
Many websites (e.g: Facebook) redirect to an error page browsers that
don't support JavaScript by using a meta redirect inside a noscript tag
like this:

```
<noscript>
    <meta http-equiv="refresh" content="0; URL=/?_fb_noscript=1"/>
</noscript>
```

This removes noscript and script tags from the source before searching
for the meta tag.


kmike:
This moves https://github.com/scrapy/scrapy/issues/18 fix to w3lib, so we may want to remove it from Scrapy. See also: https://github.com/scrapy/scrapy/issues/1422.

PR #40, opened 2015-10-29
Fix uncaught UnicodeDecodeError when decoding entities (Thanks @horva!)

Youwotma:
Only happens for Windows-1252 control entities in the range 0x80-0x9f


Digenis:
How about dealing explicitly with the 5 undefined characters instead of catching a ValueError which may hide other bugs?

Youwotma:
Because you could follow the same reasoning for the other statement in the `try`, but this time there are thousands of undefined Unicode code points.

kmike:
I was also about to suggest the same as @Digenis, but it is easier to make a mistake if we specify code points ourselves.

Digenis:
I had suggested it because they are just 5.

PR #41, opened 2016-01-25
[MRG+1] Fix typo 'zip_zafe' in setup.py

nyov:
Squashing the following `UserWarning`:
```
/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'zip_zafe'
  warnings.warn(msg)
```

(...unless this option can/should actually be removed? Seems like it wasn't active until now.
Been there since https://github.com/scrapy/w3lib/commit/ac08e )

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ setup(
     url='https://github.com/scrapy/w3lib',
     packages=find_packages(exclude=('tests', 'tests.*')),
     include_package_data=True,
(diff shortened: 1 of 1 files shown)


kmike:
A good catch. I think it is better to have this option enabled - it means less work for setuptools at install time, it won't have to figure out if there are non-python files in distribution.

PR #44, opened 2016-02-29
Fix get_base_url for non-ascii urls for Python 3

lopuhin:
I added tests by @redapple, and fixed ``get_base_url`` to do urljoin on unicode strings.

The line with ``baseurl = str_to_unicode(baseurl, encoding)`` is required because ``get_base_url`` seems to be expected to work with byte urls as well (``test_get_base_url`` in the same file). I am not sure which encoding to use here, decided to be more permissive to avoid breaking existing python 2 code which might be using utf-8 encoded strings here.

Related to https://github.com/scrapy/scrapy/issues/1783

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -286,6 +286,28 @@ class GetBaseUrlTest(unittest.TestCase):
             </html>"""
         self.assertEqual(get_base_url(text, baseurl), 'https://example.org')
 
+    def test_get_base_url_utf8(self):
+        baseurl = u'https://example.org'
+
+        text = u"""
+            <html>
+            <head><title>Dummy</title><base href='http://example.org/snowman\u2368' /></head>
+            <body>blahablsdfsal&amp;</body>
+            </html>"""
+        self.assertEqual(get_base_url(text, baseurl),
+                         'http://example.org/snowman%E2%8D%A8')
+
+    def test_get_base_url_latin1(self):
+        baseurl = u'https://example.org'
+
+        text = u"""
+            <html>
+            <head><title>Dummy</title><base href='http://example.org/sterling\u00a3' /></head>
+            <body>blahablsdfsal&amp;</body>
+            </html>"""
+        self.assertEqual(get_base_url(text, baseurl, encoding='latin-1'),
+                         'http://example.org/sterling%A3')
+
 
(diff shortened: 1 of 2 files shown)


redapple:
@lopuhin ,
after testing a few cases with Chrome and Firefox, and also discussing with @kmike ,
I'm having second thoughts on tests expecting non-UTF8 encoded data prior to percent-escaping in URL paths.

What I'm seeing with modern browsers is that the encoding of the page (or of a `<form>` within the page) only affects the query part of URLs, not paths.
So I believe base URL (paths) should always be crafted with UTF-8 encoded characters, then percent-escaped.
Hence,
`u'http://example.org/sterling\u00a3'` should translate to `b'http://example.org/sterling%C2%A3'`

Youwotma:
if I'm reading the spec correctly, compliant web servers should interpret encoded and decoded paths the same, but in practice not all of them do (e.g. [this](http://www.quadratec.com/vendor/G2+Axle+%26+Gear/index.htm) vs [this](http://www.quadratec.com/vendor/G2+Axle+&+Gear/index.htm))

redapple:
@Youwotma , true, [RFC 3986, §3.3 Path](https://tools.ietf.org/html/rfc3986#section-3.3) does allow `"&"` in path segments:

```
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```
This server seems to treat `"&"` as special within the path component, [while it shouldn't](https://tools.ietf.org/html/rfc3986#section-2.2):

> URI producing applications should percent-encode data octets that
   correspond to characters in the reserved set **unless these characters
   are specifically allowed by the URI scheme to represent data in that
   component**.  If a reserved character is found in a URI component and
   no delimiting role is known for that character, then it must be
   interpreted as representing the data octet corresponding to that
   character's encoding in US-ASCII.

redapple:
@lopuhin , I submitted a new PR continuing your work and following https://github.com/scrapy/w3lib/pull/44#issuecomment-197789555
would you mind having a look?

lopuhin:
@redapple #45 looks excellent! Closing this

PR #45, opened 2016-03-22
Fix tests on non-ASCII characters in URL + new safe_url_string()

redapple:
This is a continuation of https://github.com/scrapy/w3lib/pull/44, related to https://github.com/scrapy/scrapy/issues/1783.

The rule I followed here is to always use UTF-8 encoding for URL path component, and `encoding` for query string or form data. (the use of UTF-8 can be overriden if needed using the new `path_encoding` argument to `safe_url_string()`)

This follows what I understand from http://tools.ietf.org/html/rfc3987#page-10 and what browser tests (Chrome and Firefox) exhibit.

This PR changes `safe_url_string()` in an incompatible way, but I believe it's more correct this way.

It also includes verbatim copies of Scrapy's `to_unicode()`, `to_bytes()` and `to_native_str()`

Note: This PR also removes `urljoin_rfc` which is deprected since v1.1

### DIFF
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -7,7 +7,7 @@ import re
 import posixpath
 import warnings
 import six
-from six.moves.urllib.parse import (urljoin, urlsplit, urlunsplit,
+from six.moves.urllib.parse import (urlsplit, urlunsplit,
                                     urldefrag, urlencode, urlparse,
                                     quote, parse_qs, parse_qsl)
 from six.moves.urllib.request import pathname2url, url2pathname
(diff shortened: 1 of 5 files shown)


w3lib/url.py
```diff
@@ -47,13 +50,13 @@ def urljoin_rfc(base, ref, encoding='utf-8'):
...
+    return urljoin(str_base, str_ref)
 
 _reserved = b';/?:@&=+$|,#' # RFC 3986 (Generic Syntax)
 _unreserved_marks = b"-_.!~*'()" # RFC 3986 sec 2.3
 _safe_chars = _ALWAYS_SAFE_BYTES + b'%' + _reserved + _unreserved_marks
 
-def safe_url_string(url, encoding='utf8'):
+def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
     """Convert the given url into a legal URL by escaping unsafe characters
     according to RFC-3986.
```
lopuhin: I think the docstring below should also be updated: the unicode url is not converted to str using given encoding. I'd suggest something like (following the summary from the PR):
```
If a bytes url is given, it is first converted to str using the given
encoding (which defaults to 'utf-8'). 'utf-8' encoding is used for
URL path component (unless overriden by path_encoding), and given
encoding is used for query string or form data.
When passing a encoding, you should use the encoding of the
original page (the page from which the url was extracted from).
```

redapple: @lopuhin , indeed. very nice docstring, thanks!

lopuhin:
@redapple I've checked all the tests, apart from the old docstring everything looks good!
I think now the only part that is different from the browser behavior is handling of unicode domains, but this is a different issue, right?

redapple:
@lopuhin , correct. I've omitted unicode domains tests.
I believe it can be built on top (I hope it would not break too much of this PR)

lopuhin:
@redapple agreed, I think it'll just mean adding more tests, all current tests will stay the same

w3lib/url.py
```diff
@@ -7,8 +7,11 @@
 import posixpath
 import warnings
 import six
-from six import moves
-from w3lib.util import unicode_to_str
+from six.moves.urllib.parse import (urljoin, urlsplit, urlunsplit,
+                                    urldefrag, urlencode, urlparse,
+                                    quote, parse_qs, parse_qsl)
+from six.moves.urllib.request import pathname2url, url2pathname
+from w3lib.util import to_bytes, to_native_str, to_unicode
```
kmike: I wonder if it is a time to add `__all__` to w3lib.url - see https://github.com/redapple/w3lib/commit/7d83b092f4054ba439a115ec041e9b190385b155 and https://github.com/scrapy/scrapy/blob/ebef6d7c6dd8922210db8a4a44f48fe27ee0cd16/scrapy/utils/url.py#L16;

redapple:
Note: `str_to_unicode` and `unicode_to_str` can be removed (but https://github.com/scrapy/w3lib/pull/46 also need to happen)

redapple:
@lopuhin , I've added support for IDNs

w3lib/url.py
```diff
@@ -7,69 +7,62 @@
...
 
 # Python 2.x urllib.always_safe become private in Python 3.x;
 # its content is copied here
 _ALWAYS_SAFE_BYTES = (b'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
                       b'abcdefghijklmnopqrstuvwxyz'
                       b'0123456789' b'_.-')
-
-
-def urljoin_rfc(base, ref, encoding='utf-8'):
```
redapple: so this backfired on Scrapinghub's testing bots.
@kmike , do you think we should bring it back? announcing it will be removed in, say, 1.16?

kmike: I'm fine with bringing it back, there is no need for aggressive removal of deprecated code.

PR #47, opened 2016-03-30
Support pre-releases

redapple:
Similar to [what Scrapy does](https://github.com/scrapy/scrapy/wiki/Scrapy-release-procedure).

The idea is to use `bumpversion minor --no-tag` after an official release so that master branch shows a newer version.
I'm not bumpversion-knowledgeable so I copied lines from scrapy's `.bumpversion.cfg`

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -4,7 +4,29 @@ commit = True
 tag = True
 tag_name = v{new_version}
 
+parse = ^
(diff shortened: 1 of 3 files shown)


redapple:
Closing for now.
It's proving hard to use in pratice.
See https://github.com/scrapy/scrapy/pull/2159.

PR #49, opened 2016-03-31
Update release notes for 1.14

redapple:


kmike:
I think it is better to bump version numbers all at once, when doing a release, and use TBA/TBD in changelog title for unreleased versions. We may also mention zip_safe fix in release notes. Other than that PR looks fine, nice to see changelog updated!

PR #50, opened 2016-04-06
Handle bytes URL with wrong encoding by percent-encoding offending bytes

redapple:
Related to https://github.com/scrapy/scrapy/pull/1906#issuecomment-206510063

Scrapy 1.0 branch [transforms Unicode URLs in `Requests` init to bytes](https://github.com/scrapy/scrapy/blob/4161a8c82c371ee34841a824ac99de9e2c8cc090/scrapy/http/request/__init__.py#L55) before passing them to `safe_url_string`. Encoding used for this should be UTF8 and not the `encoding` passed at `Request` init.

Example: `(u"http://www.scrapy.org/price/\xa3", encoding='latin1')` becomes `b'http://www.scrapy.org/price/\xa3'` for `safe_url_string` with encoding information (latin1) lost on the way.
Before this patch, UTF-8 is assumed, which fails when converting back to Unicode.

To handle legacy pre-scrapy1.1 code, with wrong passed (or default) encoding, this patch fallbacks to percent-encoding of the bytes that the codec can't handle.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -106,6 +106,19 @@ class UrlTests(unittest.TestCase):
         self.assertTrue(isinstance(safeurl, str))
         self.assertEqual(safeurl, "http://www.example.com/%C2%A3?unit=%B5")
 
+    def test_safe_url_string_bytes_input_nonutf8(self):
+        # latin1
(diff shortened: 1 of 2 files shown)


redapple:
Sorry, I pushed to "master" by accident.
@kmike , would you agree with the change? (I can revert if needed)

PR #55, opened 2016-04-10
Add __all__ to w3lib.url package

### LINKED ISSUE #54: w3lib 1.14 breaks scrapy 0.24.6 usage of urlparse.urldefrag
```shell
$ pip install scrapy==0.24.6
```
This currently pulls in w3lib 1.14.0 or 1.14.1.

```shell
$ python -c "from scrapy import Request; req = Request(url='http://www.example.com')"
```

Gives:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/avish/.virtualenvs/w3lib-bug/local/lib/python2.7/site-packages/scrapy/http/request/__init__.py", line 26, in __init__
    self._set_url(url)
  File "/home/avish/.virtualenvs/w3lib-bug/local/lib/python2.7/site-packages/scrapy/http/request/__init__.py", line 52, in _set_url
    self._url = escape_ajax(safe_url_string(url))
  File "/home/avish/.virtualenvs/w3lib-bug/local/lib/python2.7/site-packages/scrapy/utils/url.py", line 102, in escape_ajax
    defrag, frag = urlparse.urldefrag(url)
AttributeError: 'function' object has no attribute 'urldefrag'
```

Downgrading to `w3lib==1.13.0` fixes this issue.

The problem doesn't happen with `scrapy>=1.0` (but 1.0 introduced breaking changes so some projects still use 0.24)

redapple:
Fixes #54

### DIFF
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -264,3 +264,8 @@ def any_to_uri(uri_or_path):
         return path_to_file_uri(uri_or_path)
     u = urlparse(uri_or_path)
     return uri_or_path if u.scheme else path_to_file_uri(uri_or_path)
+
+
+__all__ = ["add_or_replace_parameter", "any_to_uri", "file_uri_to_path",
+           "path_to_file_uri", "safe_download_url", "safe_url_string",
+           "url_query_cleaner"]
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -267,5 +267,8 @@ def any_to_uri(uri_or_path):
 
 
 __all__ = ["add_or_replace_parameter", "any_to_uri", "file_uri_to_path",
-           "path_to_file_uri", "safe_download_url", "safe_url_string",
-           "url_query_cleaner"]
+           "is_url", "path_to_file_uri", "safe_download_url",
+           "safe_url_string", "url_query_cleaner", "url_query_parameter",
+
+           # this last one is deprecated ; include it to be on the safe side
+           "urljoin_rfc"]


w3lib/url.py
```diff
@@ -264,3 +264,8 @@ def any_to_uri(uri_or_path):
         return path_to_file_uri(uri_or_path)
     u = urlparse(uri_or_path)
     return uri_or_path if u.scheme else path_to_file_uri(uri_or_path)
+
+
+__all__ = ["add_or_replace_parameter", "any_to_uri", "file_uri_to_path",
```
kmike: what do you think about adding deprecated and semi-private function here as well?
`urljoin_rfc`, `is_url`, `url_query_parameter` are missing.

redapple: * `url_query_parameter`: my bad, I missed this one when looking at https://w3lib.readthedocs.org/en/latest/w3lib.html#module-w3lib.url
* `is_url` is used in scrapy even, so makes sense to add, although it's not documented.
* I'm not sure about `urljoin_rfc`, I'd prefer not, even if it's documented.

any other you had in mind?

redapple: (ok, seems I did not push latest local changes)

kmike: Exposing variables like `_safe_chars` could have made sense, but `import *` doesn't import them anyways, so no need to add them to `__all__`.

What's wrong with exposing `urljoin_rfc`? Not exposing it may break some code. The whole `__all__` fix is needed for backwards compatibility, so why fix it for some functions but not with others?

kmike: You already fixed that :)

redapple: Nothing wrong, just that "exposing" is explicitly make keep people using it while we don't recommend it.
I added it anyway

PR #56, opened 2016-04-11
Add Python 3.5 env for tox and Travis CI

redapple:


.travis.yml
```diff
@@ -6,6 +6,7 @@ env:
   - TOXENV=pypy
   - TOXENV=py33
   - TOXENV=py34
+  - TOXENV=py35
```
kmike: To make it work use python:3.5 instead of python:2.7, Travis will download Python 3.5 in this case.

redapple: I got too confident :)

PR #61, opened 2016-07-07
Type hints

kmike:
I've tried to add [PEP-484](https://www.python.org/dev/peps/pep-0484/) type hints in this PR (checked using mypy). They showed how tricky are our string types :) For each string there are several options:
- str
- bytes
- typing.AnyStr
- typing.Text (or six.text_type)
- w3lib._types.String (a custom type I had to add)

There are several interesting points revealed by these type checks, e.g. that `html_body_declared_encoding` only worked in Python 3 because stdlib `encodings.normalize_encoding` accept binary values (I'm not sure this stdlib module is even documented).

---

Update: this PR is rather useless for the end users because of https://github.com/python/mypy/issues/1190; it also adds `typing` runtime dependency which could be good to avoid. Switch to stub files?

### DIFF
--- a/w3lib/_types.py
+++ b/w3lib/_types.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+"""
+Which string type to use?
+=========================
+
+1. Variable is an URL ==> use ``str``
+2. Variable is binary; unicode is not accepted ==> use ``bytes``
+3. Variable is text, and it can be only unicode in Python 2 ==> use
+   ``six.text_type``  (or typing.Text??)
+4. Variable is text, but it can be str in Python 2 ==> use w3lib._types.String
+5. Variable can be either bytes or unicode both in Python 2
+   and Python 3 ==> use typing.AnyStr
+6. Variable should be str (==bytes) in Python 2
+   and str (==unicode) in Python 3 ==> use ``str``.
+
+"""
+
+from __future__ import absolute_import
+from typing import Union
+import six
+
+if six.PY2:
+    String = Union[bytes, unicode]
+else:
+    String = str
--- a/w3lib/_types.py
+++ b/w3lib/_types.py
@@ -7,14 +7,14 @@ Which string type to use?
 2. Variable is binary; unicode is not accepted ==> use ``bytes``
 3. Variable is text, and it can be only unicode in Python 2 ==> use
    ``six.text_type``  (or typing.Text??)
-4. Variable is text, but it can be str in Python 2 ==> use w3lib._types.String
+4. Variable is text, but it can be ascii or utf8-encoded str
+   in Python 2 ==> use w3lib._types.String
 5. Variable can be either bytes or unicode both in Python 2
    and Python 3 ==> use typing.AnyStr
 6. Variable should be str (==bytes) in Python 2
    and str (==unicode) in Python 3 ==> use ``str``.
 
 """
-
 from __future__ import absolute_import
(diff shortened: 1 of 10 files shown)


w3lib/html.py
```diff
@@ -7,16 +7,19 @@
 import re
 import six
 from six import moves
+from typing import AnyStr, Optional, Iterable, Tuple, Union, Sequence
```
kmike: Get off annoying bot, these imports are used. You're not aware of PEP-484.

w3lib/encoding.py
```diff
@@ -143,21 +160,23 @@ def resolve_encoding(encoding_alias):
 ]
 _FIRST_CHARS = set(c[0] for (c, _) in _BOM_TABLE)
 
+
 def read_bom(data):
+    # type: (bytes) -> Tuple[str, bytes]
     r"""Read the byte order mark in the text, if present, and
     return the encoding represented by the BOM and the BOM.
 
     If no BOM can be detected, ``(None, None)`` is returned.
```
lopuhin: A nitpick - this is not reflected by the type

kmike: Yeah; currently mypy assumes that None is a valid value for any type, so `Tuple[str, bytes]` allows `None, None`. I recall using `Union[Tuple[str, bytes], Tuple[None, None]]` in some other place, but I'm not sure it works.

lopuhin: Ah, didn't know that, thanks! mypy behavior makes sense for tuple, although not that I would expect it in general :)

kmike: They have docs for this: http://mypy.readthedocs.io/en/latest/planned_features.html. Let me try Union again :)

kmike: it returns a weird error for this line if annotation is changed to Union:

```bom_enc, bom = read_bom(html_body_str)```

> w3lib/encoding.py: note: In function "html_to_unicode":
w3lib/encoding.py:274: error: 'Union[Tuple[builtins.str, builtins.bytes], Tuple[void, void]]' object is not iterable

It looks related to Union handling, not to Tuple[None, None] handling because it also fails if Tuple[str, str] is used as a second option for Union.

Comment (lopuhin) [inline w3lib/encoding.py]: So mypy does not see that all union "kinds" here are iterable, that's a pity.

File: w3lib/_types.py [hunk 3]
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+"""
+Which string type to use?
```
eliasdorneles: :heart:

w3lib/encoding.py
kmike: relevant mypy issue: https://github.com/python/mypy/issues/1575

eliasdorneles:
> it also adds typing runtime dependency which could be good to avoid

if we want to avoid the runtime dependency, we could use a hack like:

```
if os.getenv('ENABLE_TYPE_CHECKS'):
    from typing import ...
```

it'd make the imports kinda ugly, tho.

lucywang000:
FTR: I'll continue this PR soon.

kmike:
FTR: There was a lot of new development in past 2 years on typing. Now type annotations from third-party libraries can be even more useful, as mypy implemented support for https://www.python.org/dev/peps/pep-0561/ - it is possible to check user code against annotations of third-party libraries.

joe733:
Bumping to close outdated PR.

PR #64, opened 2016-07-14
Update release notes for 1.14.3

redapple:


kmike:
Thanks @redapple for taking care of releases!

PR #68, opened 2016-07-29
Run doctests with tox (and py.test)

redapple:


redapple:
@kmike , are you ok with this?

kmike:
👍

conftest.py
```diff
@@ -1,7 +1 @@
-import six
-collect_ignore = []
-
-if six.PY3:
-    for fn in open('tests/py3-ignores.txt'):
```
eliasdorneles: did this fix all PY3 tests?
should the py3-ignores.txt file be removed?

redapple: ah, damn, I knew I had forgotten something. I was meant to remove that file ealier.
Only `urljoin_rfc` doctest fails with Python 3 (as it's deprecated, I didn't dig and skipped it)

PR #71, opened 2016-08-12
[MRG+1] data URI parser.

ArturGaspar:


redapple:
I wonder if this shouldn't be in `w3lib.encoding`

ArturGaspar:
I don't see why `encoding`.

I think it could belong in `w3lib.url`.

redapple:
Right, `w3lib.url` looks good for this

redapple:
LGTM. Thanks @ArturGaspar !
cc @kmike , @eliasdorneles

ArturGaspar:
Only fixed an invalid URL used in the tests above.

w3lib/url.py
```diff
@@ -276,11 +277,96 @@ def any_to_uri(uri_or_path):
...
+        else:
+            break
+
+    is_base64, data = uri.split(b',', 1)
+    if is_base64:
+        if is_base64 != b";base64":
+            raise ValueError("invalid data URI")
+        data = base64.b64decode(data)
+
+    return media_type, media_type_params, data
```
eliasdorneles: I think it would be better to return a dictionary instead, to use future maintenance and also to make the result immediately understood by the occasional user.

w3lib/url.py
```diff
@@ -276,11 +277,96 @@ def any_to_uri(uri_or_path):
...
+
+    media_type = "text/plain"
+    media_type_params = {}
+
+    m = _mediatype_pattern.match(uri)
+    if m:
+        media_type = m.group().decode()
+        uri = uri[m.end():]
+    else:
+        media_type_params['charset'] = "US-ASCII"
```
eliasdorneles: If I understood correctly (reading the wikipedia and skimming over the RFC), I understand that `US-ASCII` is the assumed encoding when no `charset` parameter is given.

If `charset` parameter is given, perhaps we should decode using that charset?
Or return raw_data plus the attempted decoded data (it could be a lazy attribute in a class)?

ArturGaspar: The interpretation of the `charset` parameter as the encoding is particular to the `text` media type.

IMO this function shouldn't try to interpret the media type, or any other aspect of the data.

w3lib/url.py
ArturGaspar: @eliasdorneles Good suggestion. I used a namedtuple like the standard library URL parsing functions.

w3lib/url.py
eliasdorneles: :+1: looks good to me

tests/test_url.py
```diff
@@ -553,6 +553,69 @@ def test_canonicalize_url_idna_exceptions(self):
...
+        result = parse_data_uri(u"data:,A%20brief%20note")
+        self.assertEqual(result.data, b"A brief note")
+
+    def test_bytes_uri(self):
+        result = parse_data_uri(b"data:,A%20brief%20note")
+        self.assertEqual(result.data, b"A brief note")
+
+    def test_non_ascii_uri(self):
+        with self.assertRaises(UnicodeEncodeError):
+            parse_data_uri(u"data:,é")
```
kmike: I think raising UnicodeEncodeError is not a good way to handle it in Python 3. We settled on unicode URLs by default for Python 3, they shouldn't raise errors. I tried it in Firefox - it escaped this URL to `data:,%C3%A9` and then processed it. `w3lib.url.safe_url_string` function escaped URL the same way. I'm not 100% sure about that, but it looks like we should use safe_url_string here instead of encoding to ascii. This way data: url will be handled the same way as a browser handles them. What do you think?

ArturGaspar: Makes sense.

redapple:
@kmike , are you ok with the latest changes?

redapple:
@kmike ping :)

tests/test_url.py
```diff
@@ -553,6 +553,69 @@ def test_canonicalize_url_idna_exceptions(self):
...
+
+    def test_wrong_base64_param(self):
+        with self.assertRaises(ValueError):
+            parse_data_uri("data:text/plain;baes64,SGVsbG8sIHdvcmxkLg%3D%3D")
+
+    def test_wrong_scheme(self):
+        with self.assertRaises(ValueError):
+            parse_data_uri("http://example.com/")
+
+
```
redapple: [Wikipedia](https://en.wikipedia.org/wiki/Data_URI_scheme#Syntax) states:
> Data URIs encoded in Base64 may contain whitespace for human readability.

@ArturGaspar , can you add tests for multiline content?

[RFC2397 has an example](https://tools.ietf.org/html/rfc2397#section-4):

```
 <IMG
   SRC="data:image/gif;base64,R0lGODdhMAAwAPAAAAAAAP///ywAAAAAMAAw
   AAAC8IyPqcvt3wCcDkiLc7C0qwyGHhSWpjQu5yqmCYsapyuvUUlvONmOZtfzgFz
   ByTB10QgxOR0TqBQejhRNzOfkVJ+5YiUqrXF5Y5lKh/DeuNcP5yLWGsEbtLiOSp
   a/TPg7JpJHxyendzWTBfX0cxOnKPjgBzi4diinWGdkF8kjdfnycQZXZeYGejmJl
   ZeGl9i2icVqaNVailT6F5iJ90m6mvuTS4OK05M0vDk0Q4XUtwvKOzrcd3iq9uis
   F81M1OIcR7lEewwcLp7tuNNkM3uNna3F2JQFo97Vriy/Xl4/f1cf5VWzXyym7PH
   hhx4dbgYKAAA7"
   ALT="Larry">
```

redapple:
LGTM, @ArturGaspar. Thanks!

tests/test_url.py
```diff
@@ -553,6 +553,86 @@ def test_canonicalize_url_idna_exceptions(self):
...
+        self.assertEqual(result.media_type, "text/plain")
+        self.assertEqual(result.data, b"Hello, world.")
+
+    def test_base64_spaces(self):
+        result = parse_data_uri("data:text/plain;base64,SGVsb%20G8sIH%0A%20%20"
+                                "dvcm%20%20%20xk%20Lg%3D%0A%3D")
+        self.assertEqual(result.media_type, "text/plain")
+        self.assertEqual(result.data, b"Hello, world.")
+
+        result = parse_data_uri("data:text/plain;base64,SGVsb G8sIH\n  "
```
ArturGaspar: This URI with actual spaces rather than escape sequences is not valid, I think it is not a good idea to test for it. The test passes because parsing just so happens to be implemented in a way that handles it as well.

RFC 3986 appendix C sort of allows for spaces, but per my understanding it would be the responsibility of the program that extracts the URI to normalise it, not of a library that is not dealing with the source of the URI directly. Also only whitespace in the base64 part is handled, not in any part of the URI.

ArturGaspar: Firefox seems to allow spaces anywhere after the scheme. Chrome seems to allow spaces only on the base64 part.

If the goal is to be compatible with browsers, then perhaps this URI should be considered valid?

redapple: imho, the implementation should support whitespace in base64 encoded data part.
I don't think we need to support whitespace elsewhere. (so that would mean following what Chrome does, I believe)

ArturGaspar: That's what it already does, then.

redapple:
@eliasdorneles , @kmike , what do you think of the patch now?

w3lib/url.py
```diff
@@ -276,11 +278,102 @@ def any_to_uri(uri_or_path):
...
+    type parameters, and data.
+
+    """
+
+    if not isinstance(uri, bytes):
+        uri = safe_url_string(uri).encode('ascii')
+
+    scheme, uri = uri.split(b':', 1)
+    if scheme.lower() != b'data':
+        raise ValueError("not a data URI")
```
kmike: What do you think about raising the same ValueError if there is no scheme at all (currently it will fail earlier, when unpacking split result)?

w3lib/url.py
```diff
@@ -276,11 +278,102 @@ def any_to_uri(uri_or_path):
...
+                value = re.sub(br'\\(.)', r'\1', value_quoted)
+            media_type_params[attribute.decode()] = value.decode()
+            uri = uri[m.end():]
+        else:
+            break
+
+    is_base64, data = uri.split(b',', 1)
+    if is_base64:
+        if is_base64 != b";base64":
+            raise ValueError("invalid data URI")
```
kmike: It'd be nice to have an error message different from the error message of other ValueError, and also to raise ValueError if uri.split returns only a single result

kmike:
Ok, sorry for the delay. Let's merge it.

kmike:
Thanks @ArturGaspar!

PR #75, opened 2016-10-05
[Python2] Handle urlparse'd bytes URLs in canonicalize_url()

redapple:
Fixes https://github.com/scrapy/scrapy/issues/2304

### DIFF
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -422,7 +422,9 @@ def parse_url(url, encoding=None):
     parsed url)
     """
     if isinstance(url, ParseResult):
-        return url
+        # we need to make sure we have unicode parts
+        # otherwise, things like IDNA encoding may fail on bytes
+        return ParseResult(*(to_unicode(part, encoding) for part in url))
     return urlparse(to_unicode(url, encoding))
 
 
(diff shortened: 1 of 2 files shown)


w3lib/url.py
```diff
@@ -422,7 +422,9 @@ def parse_url(url, encoding=None):
     parsed url)
     """
     if isinstance(url, ParseResult):
-        return url
+        # we need to make sure we have unicode parts
+        # otherwise, things like IDNA encoding may fail on bytes
+        return ParseResult(*(to_unicode(part, encoding) for part in url))
```
kmike: I wonder if we can remove the performance penalty from Python 3 code (if there is a penalty). 

Also, there is an issue with encodings - this assumes utf8 everywhere which may not be the case. Why is Scrapy urlparse-ing bytes urls? Should it use safe_ParseResult instead?

kmike: It seems the root issue is that IDNA encoding is defined only for unicode, not for bytes. So we should convert netloc to unicode before encoding it in _safe_ParseResult.

redapple: @kmike , regarding your first comment, scrapy is using [`to_native_string()` in the default lxml-based link extractor](https://github.com/scrapy/scrapy/blob/129421c7e31b89b9b0f9c5f7d8ae59e47df36091/scrapy/linkextractors/lxmlhtml.py#L59), which is not correct, and for which there's an open PR https://github.com/scrapy/scrapy/pull/1949

Gallaecio:
Maybe we should close this? #2304 seems to have been fixed already.

Gallaecio:
No longer applies, Scrapy dropped Python 2 support.

PR #78, opened 2016-11-03
[MRG+1] Remove trailing ":" from URL netlocs

redapple:
Fixes https://github.com/scrapy/scrapy/issues/2377


kmike:
Looks good 👍

kmike:
Reference: RFC3986, p3.2: 

URI producers **and normalizers** should omit the ":" delimiter that separates host from port if the port component is empty.

Issue #84, opened 2017-01-30
headers_raw_to_dict, problem with multiple headers with same name 


dmippolitov:
Hello. I have some problems with headers_raw_to_dict.
As you can see in a code snippet we have two X-Robots-Tag in raw headers, but i have only one in resulting dict. Thanks for your attention
```
>>> with open('../sandbox/response_headers', 'rb') as f:
...    headers = f.read()
... 
>>> headers
b'Accept-Ranges: bytes\r\nEtag: "23b8cacf-1635-546b539ee0f1a"\r\nContent-Type: text/html; charset=ISO-8859-1\r\nLast-Modified: Sun, 22 Jan 2017 21:04:18 GMT\r\nX-Robots-Tag: noindex, nofollow\r\nX-Robots-Tag: otherbot: index, nofollow\r\nServer: Apache\r\nDate: Mon, 30 Jan 2017 20:43:35 GMT\r\nContent-Language: it-IT\r\n'

>>> from w3lib.http import headers_raw_to_dict
>>> import pprint

>>> headers_dict = headers_raw_to_dict(headers)
>>> pprint.pprint(headers_dict)
{b'Accept-Ranges': [b'bytes'],
 b'Content-Language': [b'it-IT'],
 b'Content-Type': [b'text/html; charset=ISO-8859-1'],
 b'Date': [b'Mon, 30 Jan 2017 20:43:35 GMT'],
 b'Etag': [b'"23b8cacf-1635-546b539ee0f1a"'],
 b'Last-Modified': [b'Sun, 22 Jan 2017 21:04:18 GMT'],
 b'Server': [b'Apache'],
 b'X-Robots-Tag': [b'otherbot: index, nofollow']}

```
The problem is here:
```
    headers = headers_raw.splitlines()
    headers_tuples = [header.split(b':', 1) for header in headers]
    return dict([
        (header_item[0].strip(), [header_item[1].strip()])
        for header_item in headers_tuples
        if len(header_item) == 2
])
```

dmippolitov:
I meet this problem in scrapy when i scrap the page and cache is hited. Headers from first run != headers from cached response. I also created pr to fix this issue: https://github.com/scrapy/w3lib/pull/85

redapple:
Fixed in #85

PR #85, opened 2017-01-30
[MRG+1] fix headers_raw_to_dict multiple headers with one name issue

dmippolitov:
https://github.com/scrapy/w3lib/issues/84


kmike:
Looks good, thanks @dmippolitov!

PR #86, opened 2017-02-08
strip_html5_whitespace

kmike:
Extracted from https://github.com/scrapy/scrapy/pull/2547.


redapple:
@kmike , will you be adding tests for this?

kmike:
@redapple there is a doctest; do you think it deserves more?

redapple:
should be ok

PR #89, opened 2017-02-17
[MRG+1] Include additional assets in source tarball

fladi:
This includes additional assets in the source tarball that are used in
distribution packages.


kmike:
Why not. Thanks @fladi!

PR #98, opened 2017-09-14
fix(url): canonicalize_url('http://您好.中国:80/') failed

hidva:
expect: `'http://xn--5usr0o.xn--fiqs8s:80/'`
actual: `'http://xn--5usr0o.xn--:80-u68dy61b/'`

### DIFF
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -19,6 +19,28 @@ from six.moves.urllib.request import pathname2url, url2pathname
 from w3lib.util import to_bytes, to_native_str, to_unicode
 
 
+def _encode_netloc(onetloc):
+    """
+    :type onetloc: unicode
+    :rtype: unicode
+    """
+    try:
+        idx = onetloc.rfind(u':')
+        if idx != -1:
+            hostname = onetloc[:idx]
+            portpart = onetloc[idx:]
+        else:
+            hostname = onetloc
+            portpart = u''
+        # assert isinstance(hostname, unicode)
+        # assert isinstance(portpart, unicode)
+        hostname = to_unicode(hostname.encode('idna'))
+        netloc = hostname + portpart
+    except UnicodeError:
+        netloc = onetloc
+    return netloc
+
+
 # error handling function for bytes-to-Unicode decoding errors with URLs
 def _quote_byte(error):
     return (to_unicode(quote(error.object[error.start:error.end])), error.end)
@@ -373,18 +395,7 @@ __all__ = ["add_or_replace_parameter",
 def _safe_ParseResult(parts, encoding='utf8', path_encoding='utf8'):
     # IDNA encoding can fail for too long labels (>63 characters)
     # or missing labels (e.g. http://.example.com)
-    try:
-        idx = parts.netloc.rfind(u':')
-        if idx != -1:
-            hostname = parts.netloc[:idx]
-            portpart = parts.netloc[idx:]
-        else:
-            hostname = parts.netloc
-            portpart = u''
-        hostname = to_unicode(hostname.encode('idna'))
-        netloc = hostname + portpart
-    except UnicodeError:
-        netloc = parts.netloc
+    netloc = _encode_netloc(parts.netloc)
 
     return (
         to_native_str(parts.scheme),
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -374,7 +374,15 @@ def _safe_ParseResult(parts, encoding='utf8', path_encoding='utf8'):
     # IDNA encoding can fail for too long labels (>63 characters)
     # or missing labels (e.g. http://.example.com)
     try:
-        netloc = parts.netloc.encode('idna')
+        idx = parts.netloc.rfind(u':')
+        if idx != -1:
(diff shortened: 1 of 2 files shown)


w3lib/url.py
```diff
@@ -374,7 +374,15 @@ def _safe_ParseResult(parts, encoding='utf8', path_encoding='utf8'):
     # IDNA encoding can fail for too long labels (>63 characters)
     # or missing labels (e.g. http://.example.com)
     try:
-        netloc = parts.netloc.encode('idna')
+        idx = parts.netloc.rfind(':')
```
pp-qq: Here may need to consider the `parts.netloc` is non-utf8 bytes

pp-qq:
ERROR:   pypy: InterpreterNotFound: pypy

redapple:
A good catch @pp-qq !
This also affects `safe_url_string()` (and would be fixed too with your change to `_safe_ParseResult()`)
This change needs an update to the tests. Could you add 1 or 2 for this case?

w3lib/url.py
```diff
@@ -374,7 +374,15 @@ def _safe_ParseResult(parts, encoding='utf8', path_encoding='utf8'):
     # IDNA encoding can fail for too long labels (>63 characters)
     # or missing labels (e.g. http://.example.com)
     try:
-        netloc = parts.netloc.encode('idna')
+        idx = parts.netloc.rfind(u':')
```
redapple: What do you think of using `.hostname` and `.port` attributes from the `SplitResult` instead of looking for a port number in netloc?

redapple:
@pp-qq , the issue with pypy on Travis is unrelated. I'm trying to fix it in #99

pp-qq:
ERROR: pypy: InterpreterNotFound: pypy
ERROR: The command "pip install -U tox twine wheel codecov" failed and exited with 2 during .

w3lib/url.py
```diff
@@ -19,6 +19,28 @@
 from w3lib.util import to_bytes, to_native_str, to_unicode
 
 
+def _encode_netloc(onetloc):
+    """
+    :type onetloc: unicode
+    :rtype: unicode
+    """
+    try:
+        idx = onetloc.rfind(u':')
```
redapple: Thanks for the tests.
My https://github.com/scrapy/w3lib/pull/98#discussion_r142931555 still holds though.
Have you considered using `SplitResult` `.hostname` and `.port` attributes?

redapple (changes requested):

w3lib/url.py
pp-qq: https://docs.python.org/2/library/urlparse.html#urlparse-result-object. no `hostname` or `port`. just `netloc`

pp-qq: Current behavior:

```py
>>> w3lib.url.safe_url_string('https://www.baIdu.com')
'https://www.baIdu.com'  # I
>>> w3lib.url.canonicalize_url('http://www.baidu.com:80000000')
'http://www.baidu.com:80000000/'
```

when `_encode_netloc()` using `SplitResult` `.hostname` and `.port` attributes:

```py
>>> w3lib.url.safe_url_string('https://www.baIdu.com')
'https://www.baidu.com'
>>> w3lib.url.canonicalize_url('http://www.baidu.com:80000000')
'http://www.baidu.com/'
```

because [python 3.5](https://docs.python.org/3.5/library/urllib.parse.html#urllib.parse.urlsplit) says:

> `hostname` is Host name (**lower case**)...

Is this what we want?

redapple: I'm fine with lowercasing the hostname. But it is indeed a change in behavior.

redapple:
```
>>> w3lib.url.canonicalize_url('http://www.baidu.com:80000000')
'http://www.baidu.com/'
```
what's the reason for the "lost" port number?

pp-qq: [python 3.5](https://docs.python.org/3.5/library/urllib.parse.html#urllib.parse.urlsplit) says: port is None when value if not valid. and the new  `_encode_netloc()`:

```py
def _encode_netloc(parts):
	try:
		hostname = to_unicode(parts.hostname.encode('idna'))
		netloc = hostname if parts.port is None else '%s:%s' % (hostname, parts.port)
	except UnicodeError:
		netloc = parts.netloc
	return netloc
```

redapple: It's a shame that `urlsplit` and `urlparse` in Python 2.7 and 3.5 simply swallow an invalid port number.
Python 3.6 does report an error:
```
$ python
Python 3.6.2 (default, Aug 24 2017, 10:48:24) 
[GCC 6.3.0 20170406] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> t = urlsplit('http://www.BaidU.com:654444')
>>> t.port
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/urllib/parse.py", line 169, in port
    raise ValueError("Port out of range 0-65535")
ValueError: Port out of range 0-65535
```

Gallaecio (approved):

joe733:
Bumping to close outdated PR.

PR #101, opened 2018-01-17
[MRG+1] Fix the CPython 3.5/3.7 crash when decoding UTF-16/32 binary blobs

sibiryakov:
The fix is essentially to skip the whole invalid sequence when handling the error. See also https://bugs.python.org/issue32583

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,22 +12,25 @@ matrix:
       env: TOXENV=py34
     - python: 3.5
       env: TOXENV=py35
+    - python: 3.5
+      env: TOXENV=pypy3
     - python: 3.6
       env: TOXENV=py36
 
 install:
   - |
       if [ "$TOXENV" = "pypy" ]; then
-        export PYENV_ROOT="$HOME/.pyenv"
-        if [ -f "$PYENV_ROOT/bin/pyenv" ]; then
-          pushd "$PYENV_ROOT" && git pull && popd
-        else
-          rm -rf "$PYENV_ROOT" && git clone --depth 1 https://github.com/yyuu/pyenv.git "$PYENV_ROOT"
-        fi
-        # get latest portable PyPy from pyenv directly (thanks to natural version sort option -V)
-        export PYPY_VERSION=`"$PYENV_ROOT/bin/pyenv" install --list |grep -o -E 'pypy2.7-portable-[0-9][\.0-9]*$' |sort -V |tail -1`
-        "$PYENV_ROOT/bin/pyenv" install --skip-existing "$PYPY_VERSION"
-        virtualenv --python="$PYENV_ROOT/versions/$PYPY_VERSION/bin/python" "$HOME/virtualenvs/$PYPY_VERSION"
+        export PYPY_VERSION="pypy-5.9-linux_x86_64-portable"
+        wget "https://bitbucket.org/squeaky/portable-pypy/downloads/${PYPY_VERSION}.tar.bz2"
+        tar -jxf ${PYPY_VERSION}.tar.bz2
+        virtualenv --python="$PYPY_VERSION/bin/pypy" "$HOME/virtualenvs/$PYPY_VERSION"
+        source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+      fi
+      if [ "$TOXENV" = "pypy3" ]; then
+        export PYPY_VERSION="pypy3.5-5.9-beta-linux_x86_64-portable"
+        wget "https://bitbucket.org/squeaky/portable-pypy/downloads/${PYPY_VERSION}.tar.bz2"
+        tar -jxf ${PYPY_VERSION}.tar.bz2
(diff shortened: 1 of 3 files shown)


kmike:
Hey @sibiryakov,

Would you mind adding a crashing example as a test case?

CI is probably failing because we haven't updated pypy setup in .travis.yml; it should be copy-pasted from https://github.com/scrapy/scrapy/blob/master/.travis.yml#L26.

sibiryakov:
@kmike it's good now!

kmike:
Awesome, thanks @sibiryakov! I can confirm the test crashes Python 3.6 before the fix.

w3lib/encoding.py
```diff
@@ -173,15 +174,15 @@ def read_bom(data):
...
+codecs.register_error('w3lib_replace', lambda exc: (u'\ufffd', exc.end))
 
 def to_unicode(data_str, encoding):
     """Convert a str object to unicode using the encoding given
 
     Characters that cannot be converted will be converted to ``\\ufffd`` (the
     unicode replacement character).
     """
-    return data_str.decode(encoding, 'w3lib_replace')
+    return data_str.decode(encoding, 'replace' if six.PY34 else 'w3lib_replace')
```
dangra: why py3.4 is an exception here?

kmike: It is a badly named (and undocumented) variable from six which means >= 3.4 (see https://github.com/benjaminp/six/issues/147).

kmike: +1 to use a more readable check

dangra: I am fine as long as we left a comment and link to that issue on what PY34 means. May be in future versions we can replace it by what six decides to use.

.travis.yml
```diff
@@ -12,22 +12,25 @@ matrix:
...
-        # get latest portable PyPy from pyenv directly (thanks to natural version sort option -V)
-        export PYPY_VERSION=`"$PYENV_ROOT/bin/pyenv" install --list |grep -o -E 'pypy2.7-portable-[0-9][\.0-9]*$' |sort -V |tail -1`
-        "$PYENV_ROOT/bin/pyenv" install --skip-existing "$PYPY_VERSION"
-        virtualenv --python="$PYENV_ROOT/versions/$PYPY_VERSION/bin/python" "$HOME/virtualenvs/$PYPY_VERSION"
+        export PYPY_VERSION="pypy-5.9-linux_x86_64-portable"
+        wget "https://bitbucket.org/squeaky/portable-pypy/downloads/${PYPY_VERSION}.tar.bz2"
+        tar -jxf ${PYPY_VERSION}.tar.bz2
+        virtualenv --python="$PYPY_VERSION/bin/pypy" "$HOME/virtualenvs/$PYPY_VERSION"
+        source "$HOME/virtualenvs/$PYPY_VERSION/bin/activate"
+      fi
```
dangra: just a nitpick: this line and the following can be a single `elif` statement.

dangra:
For another PR, we can try to replace all the scripts related to installing pypy by pyenv shipped with ["new" trusty images at travis-ci](https://docs.travis-ci.com/user/reference/trusty/#Python-images)

a bit less fancy than using builtin pyenv is https://github.com/praekeltfoundation/travis-pyenv

tests/test_encoding.py
```diff
@@ -220,6 +220,18 @@ def test_utf16_32(self):
         self._assert_encoding('utf-16', u"hi".encode('utf-16-be'), 'utf-16-be', u"hi")
         self._assert_encoding('utf-32', u"hi".encode('utf-32-be'), 'utf-32-be', u"hi")
 
+    def test_python_crash(self):
+        import random
+        from io import BytesIO
+        random.seed(42)
+        buf = BytesIO()
+        for i in range(150000):
+            buf.write(bytes([random.randint(0, 255)]))
```
dangra: is it only me who worries about the randomness of this test case?
it looks like an use case for well known broken byte sequences or at the other end for [hypothesis](http://hypothesis.works/)

lopuhin:
> For another PR, we can try to replace all the scripts related to installing pypy by pyenv shipped with "new" trusty images at travis-ci

@dangra if I recall correctly, for scrapy we stopped using pyenv because pyenv dropped support for portable pypy binaries, and we need portable ones because ssl didn't work on default binary pypy on travis due to some binary incompatability. But it might work for w3lib.

tests/test_encoding.py
sibiryakov: It crashes when there are errors during decoding and sequence is above certain length (around 150kb). So random byte sequence is guaranteed to contain such errors. At the same time the random generator is initialised to a specific value, so it will generate the same sequence on every run.

w3lib/encoding.py
sibiryakov: The 'replace' error handling was fixed in 3.3, so that's the reason I added this check. Now I'm checking version_info directly.

PR #105, opened 2018-03-03
filter tel scheme issue #1381

nctl144:
Hey guys, I made this PR to solve the issue from scrapy/scrapy#1381. I thought it was a good idea to filter out the tel scheme if there's any in the url path since theres no information we can get from that url. What do you guys think? :)


kmike:
Hey @nctl144! I agree that we should fix it somehow, but stripping tel: from path looks wrong, we should fix the issue at earlier stages, likely at urljoin. So I'm closing this PR, but I think fixing it is still important.

PR #111, opened 2018-07-27
Fix a bug for header field getting malformed when it extends over multiple lines

okomestudio:
The message headers section of RFC 2616 (Hypertext Transfer Protocol -- HTTP/1.1) states that 

> Header fields can be extended over multiple lines by preceding each extra line with at least one SP or HT.

The current implementation of `s3lib.http.headers_raw_to_dict` function unfortunately tends to read only the first line of such multi-line header field (and discard the rest in the following loop making tuples).

This PR refactors the function to be able to properly read multi-line header field. To do this reliably, it also more strictly interprets CRLF (i.e., `\r\n`) as the end-of-line marker, as per RFC 2616.


okomestudio:
I think the CI failed because the version of `tox` used no longer support Python 3.3. Making the tests pass requires downgrade to `tox<=2.9.1`, but I do not have control over it.

w3lib/http.py
```diff
@@ -27,7 +27,17 @@ def headers_raw_to_dict(headers_raw):
 
     if headers_raw is None:
         return None
-    headers = headers_raw.splitlines()
+
+    headers = []
+    for line in headers_raw.split(b'\r\n'):
```
Gallaecio: Why limit line boundaries to `\r\n`?

okomestudio: The reason is that the specification (RFC 2616 cited above or RFC7230 which appears more recent) states that in HTTP messages, lines are folded specifically by CRLF (`\r\n`) and not by LF (`\n`). Not following this had lead to an issue where a header field value containing a LF (`\n`) was accidentally interpreted as a multiline field value. Without leading at least one space or tab char in the following line, however, the header ended up being interpreted as malformed and the lines were simply discarded when parsed out.

This is rather pathological and use of LF within a field value feels like a bad idea, but it doesn't appear to violate the specification so some HTTP messages with that structure exist.

Gallaecio: https://stackoverflow.com/a/6324184 suggests that `\n` should also be interpreted as a line terminator.

Gallaecio: I would leave `.splitlines()` nonetheless, though, given the current information we have. If later it turns out by allowing certain forms of line break we are misinterpreting header values, then we can change this accordingly.

Gallaecio:
Will you consider keeping `splitlines`?

okomestudio:
Before fixing, let me do a couple tests to see if reverting to `splitlines` do not introduce any other issues.

One thing though is that by not sticking to the standard, the code would be lenient on using `\n` as the line delimiter but it will not parse standard-compliant messages correctly.

It's okay by me if that's the design compromise for the library (and I myself am not sure how often servers do break the rule, using `\n` instead of `\r\n`). But I remember forking the repo to generate this patch due to an issue I had with real response headers being truncated. So there definitely is a trade-off, albeit it might be rare.

(One possible workaround might be to inspect each header to see which of `\n` and `\r\n` is used as the line delimiter. I'm not sure that's desired or reliable.)

w3lib/http.py
okomestudio: I propose that the function `headers_raw_to_dict` take in an optional parameter `strict` with the default value that does not change the existing behavior. This way, the code works as is, but would be ready for those who want a stricter interpretation of the HTTP standard via a simple switch.

I think exposing this switch to the users requires a similar change at the user-facing API, but that should be done in a separate PR, if desired.

okomestudio:
Sorry for taking a while but I finally managed to update the PR as requested. Please note my comment about the optional switch. I hope that the PR makes sense now.

w3lib/http.py
```diff
@@ -27,7 +33,20 @@ def headers_raw_to_dict(headers_raw):
...
     if headers_raw is None:
         return None
-    headers = headers_raw.splitlines()
+
+    if strict:
+        headers = []
+        for line in headers_raw.split(b'\r\n'):
+            if line.startswith(b' ') or line.startswith(b'\t'):
+                try:
+                    headers[-1] += (b'\r\n' + line)
```
Gallaecio: :lipstick: 
```suggestion
                    headers[-1] += b'\r\n' + line
```

Gallaecio (approved):

joe733:
Bumping to close outdated PR.

PR #112, opened 2018-09-07
Improve url_query_cleaner handling of empty querystring items.

### LINKED ISSUE #109: url_query_cleaner appends ? to urls without a query string
```python
>>> url_query_cleaner('http://domain.tld/', ['bla'], remove=True)
'http://domain.tld/?'
```

This is the code which does it: `url = '?'.join([base, sep.join(querylist)]) if querylist else base` (`querylist==['']`).

wRAR:
Fix #109.


kmike:
Looks good, thanks!

PR #118, opened 2018-12-16
[MRG+1] Added 'encoding' and 'path_encoding' parameters in 'safe_download_url'

guillaume-humbert:
Added 'encoding' and 'path_encoding' parameters in 'safe_download_url'. Make sure we can pass bytes to 'safe_download_url'.

The motivation is to be able to use 'safe_download_url' in scrapy, see this pull request comment here: https://github.com/scrapy/scrapy/pull/3494#issuecomment-440783195


kmike:
Looks good, thanks @guillaume-humbert!

PR #119, opened 2018-12-21
[MRG+1] If user doesn't want to encode unsafe characters.

maramsumanth:
Please refer issue https://github.com/scrapy/scrapy/issues/833 and PR https://github.com/scrapy/scrapy/pull/3542.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -62,6 +62,17 @@ class UrlTests(unittest.TestCase):
     def test_safe_url_string_unsafe_chars(self):
         safeurl = safe_url_string(r"http://localhost:8001/unwise{,},|,\,^,[,],`?|=[]&[]=|")
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
+        
+    def test_safe_url_string_quote_path(self):
+        safeurl = safe_url_string(u'http://google.com/"hello"', self.encoding, True)
+        self.assertEqual(safeurl, u'http://google.com/%22hello%22')
+        
+        safeurl = safe_url_string(u'http://google.com/"hello"', self.encoding, False)
+        self.assertEqual(safeurl, u'http://google.com/"hello"')
+        
+        safeurl = safe_url_string(u'http://google.com/"hello"', self.encoding)
+        self.assertEqual(safeurl, u'http://google.com/"hello"')
+        
 
     def test_safe_url_string_with_query(self):
         safeurl = safe_url_string(u"http://www.example.com/£?unit=µ")
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -64,13 +64,13 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
         
     def test_safe_url_string_quote_path(self):
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding, quote_path=True)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf-8', path_encoding='utf-8', quote_path=True)
         self.assertEqual(safeurl, u'http://google.com/%22hello%22')
         
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding, quote_path=False)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf-8', path_encoding='utf-8', quote_path=False)
         self.assertEqual(safeurl, u'http://google.com/"hello"')
         
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf-8', path_encoding='utf-8')
         self.assertEqual(safeurl, u'http://google.com/%22hello%22')
         
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -64,13 +64,13 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
         
     def test_safe_url_string_quote_path(self):
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf8',path_encoding='utf8', True)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding, quote_path=True)
         self.assertEqual(safeurl, u'http://google.com/%22hello%22')
         
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf8',path_encoding='utf8', False)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding, quote_path=False)
         self.assertEqual(safeurl, u'http://google.com/"hello"')
         
-        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf8',path_encoding='utf8')
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding=self.encoding, path_encoding=self.path_encoding)
         self.assertEqual(safeurl, u'http://google.com/%22hello%22')
         
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -64,14 +64,14 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
         
     def test_safe_url_string_quote_path(self):
-        safeurl = safe_url_string(u'http://google.com/"hello"', self.encoding, True)
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf8',path_encoding='utf8', True)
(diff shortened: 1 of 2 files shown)


maramsumanth:
@lopuhin please review!

w3lib/url.py
```diff
@@ -68,18 +68,34 @@ def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
...
-        # default encoding for path component SHOULD be UTF-8
-        quote(to_bytes(parts.path, path_encoding), _safe_chars),
-
-        # encoding of query and fragment follows page encoding
-        # or form-charset (if known and passed)
-        quote(to_bytes(parts.query, encoding), _safe_chars),
-        quote(to_bytes(parts.fragment, encoding), _safe_chars),
-    ))
+    
+    if bool(escape_URL):
```
lopuhin: It's fine to use just ``escape_URL`` instead of ``bool(escape_URL)``.

w3lib/url.py
```diff
@@ -34,7 +34,7 @@ def _quote_byte(error):
 
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
-def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
+def safe_url_string(url, encoding='utf8', escape_URL=None, path_encoding='utf8'):
```
lopuhin: I think ``escape_url`` would be a more conventional name, I think most python APIs don't uppercase it.

w3lib/url.py
```diff
@@ -68,18 +68,34 @@ def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
...
+            # default encoding for path component SHOULD be UTF-8
+            to_native_str(parts.path),
+
+            # encoding of query and fragment follows page encoding
+            # or form-charset (if known and passed)
+            quote(to_bytes(parts.query, encoding), _safe_chars),
+            quote(to_bytes(parts.fragment, encoding), _safe_chars),
+        ))
+      
+    else:
```
lopuhin: It would be great to structure this code without duplication - if we want to change something, we'd like to do it in one place.

w3lib/url.py
```diff
@@ -68,18 +68,34 @@ def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
...
-        quote(to_bytes(parts.fragment, encoding), _safe_chars),
-    ))
+    
+    if bool(escape_URL):
+        return urlunsplit((
+            to_native_str(parts.scheme),
+            to_native_str(netloc).rstrip(':'),
+
+            # default encoding for path component SHOULD be UTF-8
+            to_native_str(parts.path),
```
lopuhin: I'm not sure if the name ``escape_URL`` matches what is happening here - if ``escape_URL`` is True, we're not quoting the path, but if it's False we do quote it. To me something like ``quote_path`` would make more sense (with appropriate inverted logic).

maramsumanth: Thanks for reviewing, would ```encode_path``` be a good name?

maramsumanth: Because we are encoding ```"``` to ```%22``` in URL

maramsumanth:
Thanks @lopuhin for reviewing. I have made the necessary changes as you suggested. I do not have much idea on how to add tests. Please review it and correct me if I am wrong somewhere. :)

w3lib/url.py
```diff
@@ -34,7 +34,7 @@ def _quote_byte(error):
 
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
-def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
+def safe_url_string(url, encoding='utf8', quote_path=None, path_encoding='utf8'):
```
lopuhin: I think boolean default value would be more clear for a boolean argument than ``None``. Also I think default value should be ``True`` in this case :)

lopuhin: Also would be nice to document the new argument in the docstring below.

tests/test_url.py
```diff
@@ -62,6 +62,17 @@ def test_safe_url_string(self):
     def test_safe_url_string_unsafe_chars(self):
         safeurl = safe_url_string(r"http://localhost:8001/unwise{,},|,\,^,[,],`?|=[]&[]=|")
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
+        
+    def test_safe_url_string_quote_path(self):
+        safeurl = safe_url_string(u'http://google.com/"hello"', self.encoding, True)
```
lopuhin: I think it would be better to use a keyword argument here for both encoding and path quoting, as done in the other tests, to make the test more readable.

maramsumanth:
@lopuhin , why it is showing incorrect?

maramsumanth:
Thanks @lopuhin , for your encouragement. I think all the tests and checks are passing. :)

tests/test_url.py
```diff
@@ -62,6 +62,17 @@ def test_safe_url_string(self):
     def test_safe_url_string_unsafe_chars(self):
         safeurl = safe_url_string(r"http://localhost:8001/unwise{,},|,\,^,[,],`?|=[]&[]=|")
         self.assertEqual(safeurl, r"http://localhost:8001/unwise%7B,%7D,|,%5C,%5E,[,],%60?|=[]&[]=|")
+        
+    def test_safe_url_string_quote_path(self):
+        safeurl = safe_url_string(u'http://google.com/"hello"', encoding='utf-8', path_encoding='utf-8', quote_path=True)
```
lopuhin: I think it's fine to drop ``encoding='utf-8', path_encoding='utf-8',`` here and below, since these are default values in any case.

w3lib/url.py
```diff
@@ -34,7 +34,7 @@ def _quote_byte(error):
 
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
-def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
+def safe_url_string(url, encoding='utf8', path_encoding='utf8', quote_path=True):
```
lopuhin: Could you please add ``quote_path`` to the docstring below?

maramsumanth:
@lopuhin , I have made the changes. If any change should I make in docstring, do let me know. Thanks:)

w3lib/url.py
```diff
@@ -34,10 +34,12 @@ def _quote_byte(error):
...
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
-def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
-    """Convert the given URL into a legal URL by escaping unsafe characters
-    according to RFC-3986.
-
+def safe_url_string(url, encoding='utf8', path_encoding='utf8', quote_path=True):
+    """When quote_path is True, this will convert the given URL into a 
+    legal URL by escaping unsafe characters according to RFC-3986.
+    If quote_path is False, this method will return url as it is.
```
lopuhin: I don't think this is correct - "If quote_path is False, this method will return url as it is." - we still do a lot of processing on the URL even if quote_path is False, because it refers only to path handling. I suggest instead describe how path is handled depending on this parameter, and put it after "path component (unless overriden by path_encoding),".

maramsumanth: @lopuhin , I have made the changes 👍

maramsumanth: @lopuhin , can you please review.

w3lib/url.py
```diff
@@ -34,13 +34,14 @@ def _quote_byte(error):
...
+def safe_url_string(url, encoding='utf8', path_encoding='utf8', quote_path=True):
+    """Convert the given URL into a legal URL by escaping unsafe characters 
     according to RFC-3986.
-
+    
     If a bytes URL is given, it is first converted to `str` using the given
     encoding (which defaults to 'utf-8'). 'utf-8' encoding is used for
-    URL path component (unless overriden by path_encoding), and given
+    URL path component (unless overriden by path_encoding) only if 
+    quote_path is True, otherwise it is not encoded, and given
```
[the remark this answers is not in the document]
maramsumanth: Thanks for helping @lopuhin , got a little bit confusion between encoding and quoting. :) 👍

maramsumanth:
I am learning from my mistakes, can you merge this @lopuhin . Thank you for your valuable suggestions. 👍

maramsumanth:
@lopuhin , can you review? :)

maramsumanth:
@lopuhin , this PR merging, can help me to solve issue https://github.com/scrapy/scrapy/issues/833, I raised a PR here https://github.com/scrapy/scrapy/pull/3542, which uses this modified method of ```safe_url_string```.

maramsumanth:
@kmike , can you review and merge this? :)

maramsumanth:
@lopuhin can you merge this? :+1:

lopuhin:
@maramsumanth we have the policy that we need two reviews before the merge for non-trivial PRs, and I'm not even a core developer of w3lib, so awaiting review from a core developer.

w3lib/url.py
```diff
@@ -73,7 +76,7 @@ def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
         to_native_str(netloc).rstrip(':'),
 
         # default encoding for path component SHOULD be UTF-8
-        quote(to_bytes(parts.path, path_encoding), _safe_chars),
+        quote(to_bytes(parts.path, path_encoding), _safe_chars) if quote_path else to_native_str(parts.path),
```
Gallaecio: Any change we can refactor things a bit to keep this like shorter?

maramsumanth: The only way I found was to write if ans else in seperate lines to make it look shorter.

Gallaecio: I don’t think that would be a bad idea. Using a regular if/else statement before the return statement, and using here the corresponding value.

maramsumanth: @Gallaecio is this fine?

PR #124, opened 2019-01-30
[MRG+1] update docs to reflect changes

lucywang000:
* pytest is used now, instead of nose
* add `form` module to the docs

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -39,7 +39,7 @@ Install
 Tests
 =====
 
-`nose`_ is the preferred way to run tests. Just run: ``nosetests`` from the
+`pytest`_ is the preferred way to run tests. Just run: ``pytest`` from the
 root directory to execute tests using the default Python interpreter.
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -39,7 +39,7 @@ Install
 Tests
 =====
 
-`nose`_ is the preferred way to run tests. Just run: ``nosetests`` from the
+`pytest`_ is the preferred way to run tests. Just run: ``pytest`` from the
```
kmike: 👍

docs/w3lib.rst
```diff
@@ -26,3 +26,9 @@ w3lib Package
 
 .. automodule:: w3lib.url
     :members:
+
+:mod:`form` Module
+-----------------
+
+.. automodule:: w3lib.form
```
kmike: I'm not sure we should document it, because the only function in this module is deprecated

lucywang000: Oh sorry that makes sense to not add it, i'll update

PR #126, opened 2019-03-27
Add parameter without removing duplicated value in query string

### LINKED ISSUE #125: Bug in NEW add_or_replace_parameter (w3lib 1.20.0)
Hi everybody,

New add_or_replace_parameter doesn't work correctly for list parameters in url e.g.:
```
>>> from w3lib.url import add_or_replace_parameter
>>> add_or_replace_parameter('http://example.com/?a=1&a=2', 'b', 1)
'http://example.com/?a=2&b=1'
```
I wait `http://example.com/?a=1&a=2&b=1` but get `http://example.com/?a=2&b=1`, also I don't have this problem with w3lib 1.19.0.

rennerocha:
This should fix #125.

### DIFF
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -204,13 +204,17 @@ def _add_or_replace_parameters(url, params):
     current_args = parse_qsl(parsed.query, keep_blank_values=True)
     new_args = []
     changing_params = set(params)
+    seen_params = set()
+
     for name, value in current_args:
         if name in params:
-            new_args.append((name, params[name]))
-            del params[name]
+            if name not in seen_params:
+                new_args.append((name, params[name]))
+            seen_params.add(name)
         elif name not in changing_params:
             new_args.append((name, value))
-    new_args.extend([(name, value) for name, value in params.items()])
+
+    new_args.extend([(name, value) for name, value in params.items() if name not in seen_params])
 
     query = urlencode(new_args)
     return urlunsplit(parsed._replace(query=query))
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -204,8 +204,14 @@ def _add_or_replace_parameters(url, params):
     parsed = urlsplit(url)
     args = parse_qsl(parsed.query, keep_blank_values=True)
 
-    new_args = OrderedDict(args)
-    new_args.update(params)
+    new_args = []
+    for name, value in args:
+        if name in params.keys():
+            new_args.append((name, params[name]))
+            del params[name]
+        else:
+            new_args.append((name, value))
+    new_args.extend([(name, value) for name, value in params.items()])
 
     query = urlencode(new_args)
     return urlunsplit(parsed._replace(query=query))
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -214,7 +214,8 @@ def _add_or_replace_parameters(url, params):
         elif name not in changing_params:
             new_args.append((name, value))
 
-    new_args.extend([(name, value) for name, value in params.items() if name not in seen_params])
+    not_modified_args = [(name, value) for name, value in params.items() if name not in seen_params]
+    new_args += not_modified_args
 
     query = urlencode(new_args)
     return urlunsplit(parsed._replace(query=query))
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -205,11 +205,12 @@ def _add_or_replace_parameters(url, params):
     args = parse_qsl(parsed.query, keep_blank_values=True)
 
     new_args = []
+    changing_params = list(params.keys())
     for name, value in args:
         if name in params.keys():
             new_args.append((name, params[name]))
             del params[name]
-        else:
+        elif name not in changing_params:
             new_args.append((name, value))
     new_args.extend([(name, value) for name, value in params.items()])
 
--- a/w3lib/url.py
+++ b/w3lib/url.py
@@ -202,17 +202,15 @@ def url_query_cleaner(url, parameterlist=(), sep='&', kvsep='=', remove=False, u
 def _add_or_replace_parameters(url, params):
     parsed = urlsplit(url)
     current_args = parse_qsl(parsed.query, keep_blank_values=True)
-    new_args = []
-    changing_params = set(params)
-    seen_params = set()
 
+    new_args = []
+    seen_params = set()
     for name, value in current_args:
-        if name in params:
-            if name not in seen_params:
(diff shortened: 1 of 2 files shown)


Gallaecio:
@rennerocha Could you add tests for the case where the key you are adding or replacing already exists among the parameters? Specially when it exists twice in the existing URL. I think only the first match is replaced, I’m not sure whether or not that is intended.

rennerocha:
You are right: `add_or_replace_parameters("http://domain/test?arg1=v1&arg2=v2&arg1=v3", {"arg1": "v4")` will return `http://domain/test?arg1=v4&arg2=v2&arg1=v3` .

Not sure what should be the correct behavior. Another option is remove the duplicates and return `http://domain/test?arg1=v4&arg2=v2`, but I believe that removing existing parameters is exactly what we want to avoid.

Any other thoughts?

Gallaecio:
Well, given the name of the function and the previous behavior, I suspect that we should remove all existing parameters matching the specified name.

rennerocha:
@Gallaecio Did the change as you suggested. The only drawback is that it will not be possible to add duplicated values using this function, as the input is a dictionary.

w3lib/url.py
```diff
@@ -204,8 +204,15 @@ def _add_or_replace_parameters(url, params):
...
-    new_args.update(params)
+    new_args = []
+    changing_params = list(params.keys())
+    for name, value in args:
+        if name in params.keys():
+            new_args.append((name, params[name]))
+            del params[name]
+        elif name not in changing_params:
+            new_args.append((name, value))
+    new_args.extend([(name, value) for name, value in params.items()])
```
ejulio: No need for list comprehension here, only `new_args.extend(params.items())` is enough.
You can use `list(params.items())` to make it explicit.
Also, `new_args += params.items()` would be more _pythonic_

w3lib/url.py
```diff
@@ -204,8 +204,15 @@ def _add_or_replace_parameters(url, params):
     parsed = urlsplit(url)
     args = parse_qsl(parsed.query, keep_blank_values=True)
 
-    new_args = OrderedDict(args)
-    new_args.update(params)
+    new_args = []
+    changing_params = list(params.keys())
```
dangra: why is it converted to a list? a set would be more correct and `.keys()` method call is not required: `changing_params = set(params)`.

w3lib/url.py
```diff
@@ -204,8 +204,15 @@ def _add_or_replace_parameters(url, params):
...
+    for name, value in args:
+        if name in params.keys():
```
dangra: no need to get the list of `.keys()`, it should be possible to test like `if name in params` if `params` is a dict.

w3lib/url.py
```diff
@@ -204,8 +204,15 @@ def _add_or_replace_parameters(url, params):
...
+            new_args.append((name, params[name]))
+            del params[name]
```
dangra: I don't think mutating `params` is a good idea, I'd prefer to flag seen params in another set and ignore them on line 215.

dangra: Something in the line of 

```python
  def _add_or_replace_parameters(url, params):
      parsed = urlsplit(url)
      args = parse_qsl(parsed.query, keep_blank_values=True)
      acc = []
      seen = set()
      for k, v in args:
          if k in params:
              if k not in seen:
                  acc.append((k, params[k]))
              seen.add(k)
          else:
              acc.append((k, v))

      acc += ((k, v) for k, v in params.items() if k not in seen)
      query = urlencode(acc)
      return urlunsplit(parsed._replace(query=query))
```

rennerocha: > I don't think mutating `params` is a good idea, I'd prefer to flag seen params in another set and ignore them on line 215.

As the input argument is a dict, it makes sense that it shouldn't be modified by the function. I included a test case to avoid this scenario as well.

w3lib/url.py
rennerocha: As I removed the side-effect over the input parameters, I need this list comprehension to filter the values that I don't want (the modified ones). Replaced `.extend` to `+=` anyway.

rennerocha:
@dangra and @ejulio I did some changes after your comments. Could you please review it again?

w3lib/url.py
```diff
@@ -199,13 +199,23 @@ def url_query_cleaner(url, parameterlist=(), sep='&', kvsep='=', remove=False, u
...
+    current_args = parse_qsl(parsed.query, keep_blank_values=True)
+    new_args = []
+    changing_params = set(params)
+    seen_params = set()
+
+    for name, value in current_args:
+        if name in params:
+            if name not in seen_params:
+                new_args.append((name, params[name]))
+            seen_params.add(name)
```
Gallaecio: :lipstick: It does not matter much, but…
```suggestion
                seen_params.add(name)
```

w3lib/url.py
```diff
@@ -199,13 +199,23 @@ def url_query_cleaner(url, parameterlist=(), sep='&', kvsep='=', remove=False, u
...
+        elif name not in changing_params:
```
Gallaecio: I may be missing something, but I think this could be replaces by an `else`, and the `changing_params` variable may be unnecessary.

dangra: `else` block should be fine per https://github.com/scrapy/w3lib/pull/126#discussion_r300164837 and 
no need for `changing_params` either

w3lib/url.py
dangra: that's good 💅 :)

Gallaecio:
@rennerocha We may have a new release soon :grinning:. Please let us know if you think you’ll have time to address https://github.com/scrapy/w3lib/pull/126#discussion_r300627723 in the following days.

rennerocha:
> @rennerocha We may have a new release soon grinning. Please let us know if you think you’ll have time to address [#126 (comment)](https://github.com/scrapy/w3lib/pull/126#discussion_r300627723) in the following days.

Hi! Yes, I will work on this comment in the next days. Hopefully on time for the next release!

rennerocha:
@Gallaecio I removed the nested conditional. I believe it is clearer now. Could you please review it again?

Gallaecio:
@dangra @ejulio Can you guys have a look at the latest changes, see if this is ready to merge?

kuza:
Hi guys, how about merge this pull and make new version?

kuza:
Hi guys @dangra @ejulio @Gallaecio, how about merge this pull?

PR #127, opened 2019-05-09
remove_tags_with_content: Do not remove tags that only start with the…

### LINKED ISSUE #114: remove_tags_with_content bug
use this 
`remove_tags_with_content("""<div class="hqimg_related"><div class="to_page"><span>热点栏目</span> <s class='hotSe'></s>""", ("s",))`
get result:
`<div class="hqimg_related"><div class="to_page">`
There may be some mistakes
then modified the expression. 
`tags = '|'.join([r'<%s\s+.*?</%s>|<%s\s*/>' % (tag, tag, tag) for tag in which_ones])`
finally,it's work fine

Gallaecio:
… specified tag names

Fixes #114


kmike:
Looks good, thanks @Gallaecio and @leonharetd!

PR #128, opened 2019-05-09
[MRG+1] fix: preservation of url encoded hash signs.

### LINKED ISSUE #91: `canonicalize_url` use of `safe_url_string` breaks when an encoded hash character is encountered
`canonicalize_url` will decode all percent encoded elements in a string.

if a hash `#` is present as a percent-encoded entity (`%23`) it will be decoded... however it shouldn't be as it is a rfc delimiter and fundamentally changes the URL structure -- turning the subsequent characters into a fragment.   one of the effects is that a canonical url with a safely encoded hash will point to another url; another is that running the canonical on the output will return a different url.

example:

	>>> import w3lib.url
	>>> url = "https://example.com/path/to/foo%20bar%3a%20biz%20%2376%2c%20bang%202017#bash"
	>>> canonical = w3lib.url.canonicalize_url(url)
	>>> print canonical
	https://example.com/path/to/foo%20bar:%20biz%20#76,%20bang%202017
	>>> canonical2 = w3lib.url.canonicalize_url(canonical)
	>>> print canonical2
	https://example.com/path/to/foo%20bar:%20biz%20

what is presented as a fragment in "canonical": `#76,%20bang%202017` is part of the valid url - not a fragment - and is discarded when `canonicalize_url` is run again.

references:
* https://github.com/scrapy/w3lib/issues/5

jvanasco:
Don't decode `%23` to `#` when it appears in a URL

Fixes #91

* tests-new: test_url.CanonicalizeUrlTest.test_preserve_nonfragment_hash
* tests-pass: py27, py36
* notes: adjustment to _safe_chars suggested by @Gallaecio

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -595,6 +595,21 @@ class CanonicalizeUrlTest(unittest.TestCase):
             "http://www.{label}.com/r%C3%A9sum%C3%A9?q=r%C3%A9sum%C3%A9".format(
                     label=u"example"*11))
 
+    def test_preserve_nonfragment_hash(self):
+        # don't decode `%23` to `#`
+        self.assertEqual(canonicalize_url("http://www.example.com/path/to/%23/foo/bar"),
+                                          "http://www.example.com/path/to/%23/foo/bar")
+        self.assertEqual(canonicalize_url("http://www.example.com/path/to/%23/foo/bar#frag"),
+                                          "http://www.example.com/path/to/%23/foo/bar")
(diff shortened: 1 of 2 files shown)


kmike:
Hey! I think canonicalize_url changes are good, but I'm not sure how it affects other functions - is it a right thing to do for safe_url_string, etc.? Could you please add tests for other affected functions as well?

Gallaecio:
@jvanasco Do you think you’ll have time to address @kmike’s feedback?

Issue #70, opened 2016-08-11
It's not a good idead to parse HTML text using regular expressions


starrify:
In [`w3lib.html`](https://github.com/scrapy/w3lib/blob/master/w3lib/html.py) regular expressions are used to parse HTML texts:

``` python
_ent_re = re.compile(r'&((?P<named>[a-z\d]+)|#(?P<dec>\d+)|#x(?P<hex>[a-f\d]+))(?P<semicolon>;?)', re.IGNORECASE)
_tag_re = re.compile(r'<[a-zA-Z\/!].*?>', re.DOTALL)
_baseurl_re = re.compile(six.u(r'<base\s[^>]*href\s*=\s*[\"\']\s*([^\"\'\s]+)\s*[\"\']'), re.I)
_meta_refresh_re = re.compile(six.u(r'<meta\s[^>]*http-equiv[^>]*refresh[^>]*content\s*=\s*(?P<quote>["\'])(?P<int>(\d*\.)?\d+)\s*;\s*url=\s*(?P<url>.*?)(?P=quote)'), re.DOTALL | re.IGNORECASE)
_cdata_re = re.compile(r'((?P<cdata_s><!\[CDATA\[)(?P<cdata_d>.*?)(?P<cdata_e>\]\]>))', re.DOTALL)
```

However this is definitely incorrect when it involves commented contents, e.g.

``` python
>>> from w3lib import html
>>> html.get_base_url("""<!-- <base href="http://example.com/" /> -->""")
'http://example.com/'
```

Introducing "heavier" utilities like `lxml` would solve this issue easily, but that might be an awful idea as `w3lib` aims to be lightweight & fast.  
Or maybe we could implement some quick parser merely for eliminating the commented parts.

Any ideas?

kmike:
@starrify I believe the goal was indeed speed; also, these regexes may take e.g. only first 4096 bytes of the page, without the rest. Ideas about a proper solution are welcome! It should

a) be almost as fast as these regexes;
b) work on arbitrarily truncated HTML files.

devspyrosv:
Just to add my 2 cents and to bump this issue, 
indeed, regex parsing of the html seems to miss commented out base tags and in some cases those commented out base tags point to different websites altogether.

botzill:
Another issue here is that it does not ignore the commented tags. For example we may have a commented base tag like:

```
 <!--<base href="http://127.0.0.1" />-->
 <base href="http://www.example.com/" />

```

[...]

PR #129, opened 2019-05-10
remove_comments: Also remove truncated comments

Gallaecio:
Fixes one of the issues reported at #70

### DIFF
--- a/w3lib/html.py
+++ b/w3lib/html.py
@@ -122,7 +122,7 @@ def replace_tags(text, token='', encoding=None):
     return _tag_re.sub(token, to_unicode(text, encoding))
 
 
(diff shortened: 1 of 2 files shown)


w3lib/html.py
```diff
@@ -122,7 +122,7 @@ def replace_tags(text, token='', encoding=None):
     return _tag_re.sub(token, to_unicode(text, encoding))
 
 
-_REMOVECOMMENTS_RE = re.compile(u'<!--.*?-->', re.DOTALL)
+_REMOVECOMMENTS_RE = re.compile(u'<!--.*?(-->|$)', re.DOTALL)
```
kmike: What do you think about using non-capturing group, to make it slightly faster?

PR #133, opened 2019-07-11
Remove tab and newlines in url.safe_url_string

elacuesta:
See https://github.com/scrapy/scrapy/issues/3855

Forgive me if this is an overly naïve approach, I must admit I'm not overly familiar with the formalities of URL definition and handling.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -58,8 +58,8 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(safeurl, "http://www.example.com/%A3")
 
         self.assertTrue(isinstance(safe_url_string(b'http://example.com/'), str))
-        
-        # Remove ASCII tab and newlines
+
+    def test_safe_url_string_remove_ascii_tab_and_newlines(self):
         self.assertEqual(safe_url_string("http://example.com/test\n.html"),
                                          "http://example.com/test.html")
         self.assertEqual(safe_url_string("http://example.com/test\t.html"),
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -58,6 +58,20 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(safeurl, "http://www.example.com/%A3")
 
         self.assertTrue(isinstance(safe_url_string(b'http://example.com/'), str))
+        
+        # Remove ASCII tab and newlines
+        self.assertEqual(safe_url_string("http://example.com/test\n.html"),
+                                         "http://example.com/test.html")
+        self.assertEqual(safe_url_string("http://example.com/test\t.html"),
+                                         "http://example.com/test.html")
+        self.assertEqual(safe_url_string("http://example.com/test\r.html"),
+                                         "http://example.com/test.html")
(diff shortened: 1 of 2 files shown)


w3lib/url.py
```diff
@@ -34,6 +34,8 @@ def _quote_byte(error):
 
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
+remove_tab_newline = re.compile(r'[\t\n\r]')  # see https://url.spec.whatwg.org/#url-parsing
```
kmike: I think it should be private. Also, this regexp can be used not only for removing tabs/newlines, but for e.g. fiinding them - what do you think about naming it e.g. _ascii_tab_newline_re, and adding a link to https://infra.spec.whatwg.org/#ascii-tab-or-newline?

w3lib/url.py
```diff
@@ -56,8 +58,8 @@ def safe_url_string(url, encoding='utf8', path_encoding='utf8'):
     #     encoded with the supplied encoding (or UTF8 by default)
     #   - if the supplied (or default) encoding chokes,
     #     percent-encode offending bytes
-    parts = urlsplit(to_unicode(url, encoding=encoding,
-                                errors='percentencode'))
+    decoded = to_unicode(url, encoding=encoding, errors='percentencode')
+    parts = urlsplit(remove_tab_newline.sub('', decoded))
```
kmike: I think this newline/tab removing should be documented in safe_url_string docstring - something about workaround to make parser more compatible with https://url.spec.whatwg.org/#url-parsing, maybe "as per https://url.spec.whatwg.org/#url-parsing, tabs and newlines are removed"

elacuesta:
@kmike Applied both suggestions, thank you!

tests/test_url.py
```diff
@@ -58,6 +58,20 @@ def test_safe_url_string(self):
         self.assertEqual(safeurl, "http://www.example.com/%A3")
 
         self.assertTrue(isinstance(safe_url_string(b'http://example.com/'), str))
+        
+        # Remove ASCII tab and newlines
```
kmike: Could you please extract these tests to a separate test method?

elacuesta: Done, sorry I missed that, thanks.

w3lib/url.py
```diff
@@ -34,9 +34,12 @@ def _quote_byte(error):
 
 _safe_chars = RFC3986_RESERVED + RFC3986_UNRESERVED + EXTRA_SAFE_CHARS + b'%'
 
+_ascii_tab_newline_re = re.compile(r'[\t\n\r]')  # see https://infra.spec.whatwg.org/#ascii-tab-or-newline
```
Gallaecio: Since we are linking to the ‘living standard’, which changes constantly, I think we should not link directly to it, but through the [Wayback Machine](https://web.archive.org/). Both here and on the docstring below.

kmike: Wayback Machine is being blocked and unblocked in Russia, so I'm fine with whatwg link :)

PR #136, opened 2019-09-17
Strip spaces in canonicalize_url

### LINKED ISSUE #132: `canonicalize_url` should not accept whitespace at the begin of an URI
Accepting whitespace in the begin of URI introduces a crawling issue. Unique URIs are generated on each depth and crawler gets in a loop. 

When a URI is extracted and begins with whitespace it is being considered a relative URI and its concatenated at the end of current URI, which leads to unique URI on each crawler depth. 

But, accordingly to RFC2396 [ Page 38], the scheme part of an URI should begin with an ALPHA character. 

> "The syntax for URI scheme has been changed to require that all schemes begin with an alpha character." 

After further investigation, the core python `urlparse` function doesn't follow RFC strictly, they consider some rules. Like on Python 3.5 they say: 

> "Changed in version 3.3: The fragment is now parsed for all URL schemes (unless allow_fragment is false), in accordance with RFC 3986. Previously, a whitelist of schemes that support fragments existed." 

However a better implementation should be in accordance with RFC 3986 (which is an update of RFC2396 mentioned before), the `urlparse` should not accept whitespace at the begin of URI. 

When compared to the with Ruby implementation of RFC 2396,  they don't allow whitespace at the begin of absolute or relative URIs and throws an exception "bad URI(is not URI?)".

In order to fix that on Scrapy level, I suggest the following. 

In `scrapy/utils/url.py` add at the begin of `canonicalize_url`:

```
if re.match('[a-zA-Z]', url[:1]) == None:
    raise ValueError('Bad URI (is not a valid URI?)')
```

An a test in `tests/test_utils_url.py`:

```
def test_rfc2396_scheme_should_begin_with_alpha(self):
    self.assertRaises(ValueError, canonicalize_url," http://www.example.com")
```

But raising an exception breaks a lot of tests and can introduce unexpected behaviors. You could simple `.strip()` url, but won't make it RFC compliant.

Gallaecio:
Fixes #132


noviluni:
This definitely works for the blank spaces, but in other cases, like if the URL starts with a colon, we will interpreting it as a relative path, but in other tools, like cURL, what you get is an error: `curl: (3) Bad URL, colon is first character` (as it could be that you were missing the schema). What do you think @Gallaecio? Should this be addressed or not?

Gallaecio:
If it does not work in a web browser (and I don’t think it does), I don’t think we need to aim to support it either.

joe733:
Incoming branch is [101 commits behind](https://github.com/scrapy/w3lib/tree/strip-spaces-in-canonicalize-url)

w3lib/url.py
```diff
@@ -451,6 +451,8 @@ def canonicalize_url(url, keep_blank_values=True, keep_fragments=False,
     # UTF-8 can handle all Unicode characters,
     # so we should be covered regarding URL normalization,
     # if not for proper URL expected by remote website.
+    if isinstance(url, six.string_types):
```
felipeboffnunes, line 454:
```suggestion
    if isinstance(url, str):
```

wRAR (approved):

kmike:
Hey! Could you please elaborate on the fix? canonicalize_url main use is to generate URL fingerprints, i.e. to compare 2 URLs. So, after this change an URL with a whitespace in front and without it would be considered the same. What's the motivation for this? 

If we decide that URL-related functions should be stripping whitespace in URLs, shouldn't we be doing it in functions like safe_download_url, etc.?

kmike:
by the way, a cool commit hash @wRAR :) 1dddddbbdbc...

Gallaecio:
> Is it the idea behind the change, what's the motivation for this?

To me it is to emulate web browser behavior, i.e. entering those 2 URLs in a browser would lead the browser to load the same URL, so it makes sense to me that they are *canonicalized* the same way.

> If we decide that URL-related functions should be stripping whitespace in URLs, shouldn't we be doing it in functions like safe_download_url, etc.?

Makes sense to me.

kmike:
> To me it is to emulate web browser behavior, i.e. entering those 2 URLs in a browser would lead the browser to load the same URL, so it makes sense to me that they are canonicalized the same way.

Our canoncalization is different though, it doesn't even guarantee that the URL will load the same afterwards. For example, order of query params may change, and empty params are removed by default, which may affect the result. I think it's an anti-pattern (or a bug) to send canonicalize_url output for downloading; this is what other w3lib.url functions are for.

kmike:
Because of that, I think we should only release this change together with changes in other functions.

felipeboffnunes:
@kmike Could you elaborate a little on which other changes would allow this to stay? I can try wrapping the edges depending on how complex this seems to be.

kmike:
@felipeboffnunes it could be as easy as calling url.strip() in safe_download_url and safe_url_string, or maybe in some other functions in w3lib.url. The challenge would be to figure out what's the effect of doing so, and what's the right thing to do.

Speaking of a right thing to do, I wonder if calling url.strip is actually right, and if we should use w3lib.html.strip_html5_whitespace instead, or maybe a right thing to do is something else. What is a whitespace, basically, what should we strip? E.g. does .strip remove unicode whitespace or not, and what's right? Another question: it seems there are some references to RFCs discussed here about whitespace not allowed in front of the URL. Are they allowed at the end of  URL? If they are, should we use lstrip instead of strip? Or maybe strip is fine? Are rules the same for whitespace in the beginning and in the end?

felipeboffnunes:
@kmike  This may seem like a joke, but I believe the right path would be to look upon the current discussions, analyze if there is at the moment any provided info that explicitly goes its way to invalidating the url.strip() approach (or w3lib.html.strip_html5_whitespace) and, if we don't find any assertions about how it may break stuff, then we just do the basic.
This works because, if this breaks stuff, people will come and call us out here  ¯\_(ツ)_/¯

Gallaecio:
According to the URL living standard, the right way to handle string stripping before URL parsing is:

```
    _ASCII_TAB_OR_NEWLINE = "\t\n\r"
    _C0_CONTROL = "".join(chr(n) for n in range(32))
    _C0_CONTROL_OR_SPACE = _C0_CONTROL + " "
    input = input.strip(_C0_CONTROL_OR_SPACE)
    input = input.translate(_ASCII_TAB_OR_NEWLINE_TRANSLATION_TABLE)
```

(extract from a Python implementation of the URL parsing and serialization algorithm from the URL living standard that I am hoping to finish by the beginning of next week).

PR #138, opened 2019-09-18
fixed url.add_or_replace_parameters to work with multiple value args

### LINKED ISSUE #137: url.add_or_replace_parameter(s) removes param values for param with multiple values
url.add_or_replace_parameter(s) removes param values for param with multiple values, and only the last value is kept, even if the param name is not included in the param params.

```
url.add_or_replace_parameter('http://example.com/?foo=1&foo=2', 'bar', '3')                 
> http://example.com/?foo=2&bar=3
```

peonone:
It fixes #137 

After this PR approved and merged, I think we can add support to allow multiple values for the new parameters.


Gallaecio:
Have you seen #126?

peonone:
Ohh, I didn't see it when I created this PR, then we can close this one I think.

PR #140, opened 2019-10-16
Enable nitpicky Sphinx warnings, fix issues and fail on new issues

Gallaecio:


tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27, pypy, py34, py35, py36, py37, pypy3
+envlist = py27, pypy, py34, py35, py36, py37, pypy3, docs
```
kmike: Could you please enable it on Travis as well?

Gallaecio:
Wow, the error seems too specific to be a false positive :thinking:

kmike:
Yeah, looks like a genuine failure

Gallaecio:
It’s a regression of https://bitbucket.org/pypy/pypy/issues/2389/different-behavior-of-bytesdecode-utf8 which @lopuhin reported to PyPy in 2016. His very example then can be used to reproduce the issue.

Gallaecio:
@kmike If you are OK with running the tests on Python 2 PyPy 7.0.0 for now, we can squash and merge this.

kmike:
Sounds good, thanks @Gallaecio!

PR #165, opened 2021-03-01
Remove test broken by Python CVE-2021-23336 fix

### LINKED ISSUE #164: test_add_or_replace_parameter fails on Python 3.6.13, 3.7.10, 3.8.8, 3.9.2 due to CVE-2021-23336 fix
In Python 3.6.13, 3.7.10, 3.8.8, and 3.9.2, `urllib.parse.parse_qsl` [no longer treats `;` as a separator](https://bugs.python.org/issue42967) by default.

```pycon
Python 3.8.8 (default, Feb 19 2021, 11:04:50) 
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib.parse
>>> urllib.parse.parse_qsl('arg1=v1;arg2=v2')
[('arg1', 'v1;arg2=v2')]
```

This causes `UrlTests.test_add_or_replace_parameter` to fail.

```
__________________________________________________________ UrlTests.test_add_or_replace_parameter ___________________________________________________________

self = <tests.test_url.UrlTests testMethod=test_add_or_replace_parameter>

    def test_add_or_replace_parameter(self):
        url = 'http://domain/test'
        self.assertEqual(add_or_replace_parameter(url, 'arg', 'v'),
                         'http://domain/test?arg=v')
        url = 'http://domain/test?arg1=v1&arg2=v2&arg3=v3'
        self.assertEqual(add_or_replace_parameter(url, 'arg4', 'v4'),
                         'http://domain/test?arg1=v1&arg2=v2&arg3=v3&arg4=v4')
        self.assertEqual(add_or_replace_parameter(url, 'arg3', 'nv3'),
                         'http://domain/test?arg1=v1&arg2=v2&arg3=nv3')
    
        url = 'http://domain/test?arg1=v1;arg2=v2'
>       self.assertEqual(add_or_replace_parameter(url, 'arg1', 'v3'),
                         'http://domain/test?arg1=v3&arg2=v2')
E       AssertionError: 'http://domain/test?arg1=v3' != 'http://domain/test?arg1=v3&arg2=v2'
E       - http://domain/test?arg1=v3
E       + http://domain/test?arg1=v3&arg2=v2
E       ?                           ++++++++

tests/test_url.py:303: AssertionError
```

andersk:
In Python 3.6.13, 3.7.10, 3.8.8, and 3.9.2, `urllib.parse.parse_qsl` [no longer treats ; as a separator by default](https://bugs.python.org/issue42967).

Fixes #164.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -299,10 +299,6 @@ class UrlTests(unittest.TestCase):
         self.assertEqual(add_or_replace_parameter(url, 'arg3', 'nv3'),
                          'http://domain/test?arg1=v1&arg2=v2&arg3=nv3')
 
-        url = 'http://domain/test?arg1=v1;arg2=v2'
-        self.assertEqual(add_or_replace_parameter(url, 'arg1', 'v3'),
-                         'http://domain/test?arg1=v3&arg2=v2')
-
         self.assertEqual(add_or_replace_parameter("http://domain/moreInfo.asp?prodID=", 'prodID', '20'),
                          'http://domain/moreInfo.asp?prodID=20')
         url = 'http://rmc-offers.co.uk/productlist.asp?BCat=2%2C60&CatID=60'


Gallaecio:
I think it would be best to keep the test with updated expectations depending on the Python version being used, and include a link to https://python-security.readthedocs.io/vuln/urllib-query-string-semicolon-separator.html in a comment.

Based on https://bugs.python.org/issue42967#msg387756:

```python
url = 'http://domain/test?arg1=v1;arg2=v2'
# https://python-security.readthedocs.io/vuln/urllib-query-string-semicolon-separator.html
if 'separator' in inspect.signature(urllib.parse.parse_qs).parameters:
    expected_url = 'http://domain/test?arg1=v3'
else:
    expected_url = 'http://domain/test?arg1=v3&arg2=v2'
self.assertEqual(add_or_replace_parameter(url, 'arg1', 'v3'), expected_url)
```

andersk:
TBH I don’t see the advantage in asserting that the insecure version is insecure, but if you do, feel free to modify.

wRAR:
I'm in favor of just removing the test, especially as we won't run it with non-patched Python versions on CI anyway. I see #166 instead moves it to an xfail test which is fine for me too.

Gallaecio:
I don’t think tests should fail if run on a Python version without the security patch. We could force an unpatched Python version in CI to make sure we test both behaviors.

I don’t have a strong opinion on it, though. I’m fine with xfail or removing.

And just to be clear, I think we should eventually extend w3lib to support the old behavior, because I expect corner cases where users may need it. Not just the new option, which allows to choose between `&` and `;`, but support for interpreting either as a parameter separator.

In fact, I expect w3lib users like Scrapy to go for the old behavior by default. I don’t think Scrapy is affected by the vulnerability, but not supporting the flexibility of the standard can potentially require a custom middleware to deal with these characters in order to crawl some websites.

wRAR:
It's an interesting point about supporting old behavior.

Gallaecio:
This was technically fixed by https://github.com/scrapy/w3lib/pull/166 with the xfail approach.

PR #166, opened 2021-03-10
[CI] Migrate CI to GitHub actions

elacuesta:
Based on [Scrapy's config](https://github.com/scrapy/scrapy/tree/954c4b48e626da2f0a50c636532a1bf9092975e4/.github/workflows).

This will need a `PYPI_TOKEN` repository secret to be set. Publishing to PyPI is working against the testing instance (see https://test.pypi.org/project/w3lib/, pushed from https://github.com/elacuesta/w3lib/runs/2078491608)


tests/test_url.py
```diff
@@ -338,6 +336,13 @@ def test_add_or_replace_parameter(self):
         self.assertEqual(add_or_replace_parameter(url, 'arg1', 'v3'),
                          'http://domain/test?arg1=v3&arg2=v2')
 
+    @pytest.mark.xfail(reason="https://github.com/scrapy/w3lib/issues/164")
+    def test_add_or_replace_parameter_fail(self):
+        self.assertEqual(
+            add_or_replace_parameter('http://domain/test?arg1=v1;arg2=v2', 'arg1', 'v3'),
+            'http://domain/test?arg1=v3&arg2=v2'
+        )
+
```
elacuesta, line 345: Just a temporary workaround to avoid having the workflows fail (#164)

.github/workflows/build.yml
```diff
@@ -1,27 +1,44 @@
...
+    branches:
+      - master
 
 jobs:
   build:
 
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, pypy3]
```
wRAR, line 18: What does this change mean in practice (some versions are dropped)?

Gallaecio: Versions in this file are indicated along job definitions under `include`.

But these versions have actually been moved to the new `tests.yml` file.

wRAR (approved):

PR #123, opened 2019-01-24
Type hints based on #61

lucywang000:
This pr is based on @kmike 's work in #61.

I wanted to set the pr target to the `type-hint` branch but since it has conflicts with master I'm not able to do that. So I have to set the pr target branch to `master`.

PR #172, opened 2021-06-09
Type annotations

lucywang000:
A new attempt after #123 .

The generated document now has type hints included as well:

Screen Shot 2021-08-09 at 3 10 24 PM

### DIFF
--- a/docs/w3lib.rst
+++ b/docs/w3lib.rst
@@ -26,3 +26,5 @@ w3lib Package
 
 .. automodule:: w3lib.url
     :members:
+
+.. autoclass:: ParseDataURIResult
--- a/run-mypy.sh
+++ b/run-mypy.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-set -e
-
-mypy --txt-report . w3lib tests
--- a/run-mypy.sh
+++ b/run-mypy.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+mypy --txt-report . w3lib tests
--- a/w3lib/http.py
+++ b/w3lib/http.py
@@ -1,7 +1,11 @@
 from base64 import urlsafe_b64encode
+from typing import Any, List, MutableMapping, Optional, AnyStr, Sequence, Union, Mapping
+from w3lib.util import to_bytes
 
+HeadersDictInput = Mapping[bytes, Union[Any, Sequence]]
+HeadersDictOutput = MutableMapping[bytes, List[bytes]]
 
-def headers_raw_to_dict(headers_raw):
+def headers_raw_to_dict(headers_raw: Optional[bytes]) -> Optional[HeadersDictOutput]:
     r"""
     Convert raw headers (single multi-line bytestring)
     to a dictionary.
@@ -30,7 +34,7 @@ def headers_raw_to_dict(headers_raw):
     headers = headers_raw.splitlines()
     headers_tuples = [header.split(b":", 1) for header in headers]
 
-    result_dict = {}
+    result_dict: HeadersDictOutput = {}
     for header_item in headers_tuples:
         if not len(header_item) == 2:
             continue
@@ -46,7 +50,7 @@ def headers_raw_to_dict(headers_raw):
     return result_dict
 
 
-def headers_dict_to_raw(headers_dict):
+def headers_dict_to_raw(headers_dict: Optional[HeadersDictInput]) -> Optional[bytes]:
     r"""
     Returns a raw HTTP headers representation of headers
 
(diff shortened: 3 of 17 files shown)


lucywang000:
@Gallaecio I think this patch is ready for a review :)

w3lib/http.py
```diff
@@ -90,10 +97,8 @@ def basic_auth_header(username, password, encoding="ISO-8859-1"):
...
 
-    auth = "%s:%s" % (username, password)
-    if not isinstance(auth, bytes):
-        # XXX: RFC 2617 doesn't define encoding, but ISO-8859-1
-        # seems to be the most widely used encoding here. See also:
-        # http://greenbytes.de/tech/webdav/draft-ietf-httpauth-basicauth-enc-latest.html
-        auth = auth.encode(encoding)
-    return b"Basic " + urlsafe_b64encode(auth)
+    auth = "%s:%s" % (to_native_str(username), to_native_str(password))
+    # XXX: RFC 2617 doesn't define encoding, but ISO-8859-1
```
wRAR, line 101: is this comment still needed? I think it was about the `encode` call which is no longer done?

run-mypy.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+mypy --txt-report . w3lib tests
```
Gallaecio, line 5: Giving `tox -e typing` can be used instead, I think it may be better to remove this script.

w3lib/http.py
Gallaecio: I think it applies as long as we accept an `encoding` parameter here with that default value. We are still using the `encoding`, only now it has moved to the `to_bytes` call below.

w3lib/url.py
```diff
@@ -94,7 +102,7 @@ def safe_url_string(url, encoding="utf8", path_encoding="utf8", quote_path=True)
     return urlunsplit(
         (
             parts.scheme,
-            netloc.rstrip(":"),
+            netloc.decode().rstrip(":"),
```
Gallaecio, line 105: I’m unsure about the approach in this function.

I assume the goal is to avoid `netloc` having a different type in different parts of the code. If so, what about, instead of making it `bytes`, make it `str` with a different variable in the `try except`?

```python
    try:
        netloc_bytes = parts.netloc.encode("idna")
    except UnicodeError:
        netloc = parts.netloc
    else:
        netloc = netloc_bytes.decode()
```

w3lib/url.py
```diff
@@ -342,12 +364,15 @@ def any_to_uri(uri_or_path):
...
+ParseDataURIResult = namedtuple(
     "ParseDataURIResult", "media_type media_type_parameters data"
 )
+ParseDataURIResult.__doc__ = "The return value type of `w3lib.url.parse_data_uri`."
 
 
-def parse_data_uri(uri):
+# If we add the return type hint sphinx would error:
+#   w3lib/url.py:docstring of w3lib.url.parse_data_uri::py:class reference target not found: w3lib.url.ParseDataURIResult
+def parse_data_uri(uri: StrOrBytes):  # type: ignore
```
Gallaecio, line 375: What about adding `ParseDataURIResult` to the documentation, then (with https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html ). If we have documentation for a function, we should have documentation for its return value type.

w3lib/util.py
```diff
@@ -1,7 +1,11 @@
 from warnings import warn
+from typing import Optional
+from w3lib._types import StrOrBytes
```
Gallaecio, line 3:
```suggestion
from typing import Optional

from w3lib._types import StrOrBytes
```

lucywang000:
@wRAR @Gallaecio Thanks for your review comments. I have addressed all your feedback.

docs/w3lib.rst
```diff
@@ -26,3 +26,5 @@ w3lib Package
 
 .. automodule:: w3lib.url
     :members:
+
+.. autoclass:: ParseDataURIResult
```
Gallaecio, line 30: https://github.com/lucywang000/w3lib/pull/1

lucywang000:
@Gallaecio I've merged your fix and made some changes to make the CI pass.

Gallaecio (approved):

lucywang000:
@Gallaecio anything else I need to do to make this merged?

Gallaecio:
I don’t think so. I’m not merging myself already just in case @wRAR wants a final look first.

wRAR (approved):
Thank you!

lucywang000:
Thanks guys!

Issue #171, opened 2021-05-13
w3lib.url.safe_url_string incorrectly encode IDNA domain with port


heroesm:
Step to reproduce:

```
>>> from w3lib.url import safe_url_string
>>> safe_url_string('http://新华网.中国')
'http://xn--xkrr14bows.xn--fiqs8s'
>>> safe_url_string('http://新华网.中国:80')
'http://xn--xkrr14bows.xn--:80-u68dy61b'
```
`safe_url_string('http://新华网.中国:80')` 
expected result:
```
'http://xn--xkrr14bows.xn--fiqs8s:80'
```
real result:
```
'http://xn--xkrr14bows.xn--:80-u68dy61b'
```

Related code:
https://github.com/scrapy/w3lib/blob/ef5c11012a4d56151eee042bea06132481f318e1/w3lib/url.py#L80

```
netloc = parts.netloc.encode('idna')
```

Maybe IDNA encoding should be done on hostname rather than netloc.

wRAR:
> Maybe IDNA encoding should be done on hostname rather than netloc.

Indeed, thank you for this finding.

https://stackoverflow.com/a/804380/113586 and its comments also suggest extracting `user@pass` from `netloc` but I'm not sure if we need this.

PR #174, opened 2021-10-03
Encode hostname directly instead of netloc

Laerte:
This fix the issue 
- #171

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1104,6 +1104,18 @@ class DataURITests(unittest.TestCase):
         result = parse_data_uri("DaTa:,A%20brief%20note")
         self.assertEqual(result.data, b"A brief note")
 
+    def test_safe_url_string_encode_idna_domain_with_port(self):
+        self.assertEqual(safe_url_string('http://新华网.中国:80'),
+                         'http://xn--xkrr14bows.xn--fiqs8s:80')
+
+    def test_safe_url_string_encode_idna_domain_with_username_password_and_port_number(self):
+        self.assertEqual(safe_url_string('ftp://admin:admin@新华网.中国:21'),
+                         'ftp://admin:admin@xn--xkrr14bows.xn--fiqs8s:21')
+
+    def test_safe_url_string_encode_idna_domain_with_username_without_password_and_port_number(self):
+        self.assertEqual(safe_url_string('ftp://admin:@新华网.中国:21'),
+                         'ftp://admin:@xn--xkrr14bows.xn--fiqs8s:21')
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1105,20 +1105,33 @@ class DataURITests(unittest.TestCase):
         self.assertEqual(result.data, b"A brief note")
 
     def test_safe_url_string_encode_idna_domain_with_port(self):
-        self.assertEqual(safe_url_string('http://新华网.中国:80'),
-                         'http://xn--xkrr14bows.xn--fiqs8s:80')
+        self.assertEqual(
+            safe_url_string("http://新华网.中国:80"), "http://xn--xkrr14bows.xn--fiqs8s:80"
+        )
 
-    def test_safe_url_string_encode_idna_domain_with_username_password_and_port_number(self):
-        self.assertEqual(safe_url_string('ftp://admin:admin@新华网.中国:21'),
-                         'ftp://admin:admin@xn--xkrr14bows.xn--fiqs8s:21')
-        self.assertEqual(safe_url_string('http://Åsa:abc123@➡.ws:81/admin'),
-                         'http://%C3%85sa:abc123@xn--hgi.ws:81/admin')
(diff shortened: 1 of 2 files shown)


w3lib/url.py
```diff
@@ -95,10 +95,26 @@ def safe_url_string(
...
     # or missing labels (e.g. http://.example.com)
     try:
-        netloc_bytes = parts.netloc.encode("idna")
+        # When we have hostname we use it instead of netloc directly
+        if hostname:
+            if isinstance(username, str) and isinstance(password, str):
+                netloc_bytes += f'{username}:{password}@'.encode("idna")
+            netloc_bytes += hostname.encode("idna")
+            if port_number:
+                netloc_bytes += f":{port_number}".encode("idna")
```
Gallaecio, line 115: You are encoding everything as `idna`, but the [StackOverflow](https://stackoverflow.com/questions/804336/best-way-to-convert-a-unicode-url-to-ascii-utf-8-percent-escaped-in-python/804380#804380) answer that @wRAR linked in the issue being addressed only does so for the `hostname`. Are you sure those other parts also need to be encoded as `idna`?

tests/test_url.py
```diff
@@ -1104,6 +1104,18 @@ def test_scheme_case_insensitive(self):
...
+        self.assertEqual(safe_url_string('http://新华网.中国:80'),
+                         'http://xn--xkrr14bows.xn--fiqs8s:80')
+
+    def test_safe_url_string_encode_idna_domain_with_username_password_and_port_number(self):
+        self.assertEqual(safe_url_string('ftp://admin:admin@新华网.中国:21'),
+                         'ftp://admin:admin@xn--xkrr14bows.xn--fiqs8s:21')
+
+    def test_safe_url_string_encode_idna_domain_with_username_without_password_and_port_number(self):
+        self.assertEqual(safe_url_string('ftp://admin:@新华网.中国:21'),
+                         'ftp://admin:@xn--xkrr14bows.xn--fiqs8s:21')
```
Gallaecio, line 1117: I believe this tests an empty password. No password would mean no colon: `…//admin@…`

w3lib/url.py
Laerte: Changed, thanks.

tests/test_url.py
Laerte: Did know about that, fixed and rename the test.

tests/test_url.py
```diff
@@ -1104,6 +1104,35 @@ def test_scheme_case_insensitive(self):
...
+            "http://jap%C3%A3o:n%C3%A3o@xn--i-7iq.ws:8000/",
+        )
+
+    def test_safe_url_string_encode_idna_domain_with_username_and_empty_password_and_port_number(
+        self,
+    ):
+        self.assertEqual(
+            safe_url_string("ftp://admin:@新华网.中国:21"),
+            "ftp://admin@xn--xkrr14bows.xn--fiqs8s:21",
+        )
```
Gallaecio, line 1134: Can you also add a test for an empty password?

I suspect it will fail, because in that case I think `part.password` is `None`, but your current code seems to ignore the user if `part.password` is not a string.

Laerte: This is already the test for empty password, i'm missing something?

w3lib/url.py
```diff
@@ -95,10 +95,29 @@ def safe_url_string(
...
+    netloc_bytes = b""
+
     # IDNA encoding can fail for too long labels (>63 characters)
     # or missing labels (e.g. http://.example.com)
     try:
-        netloc_bytes = parts.netloc.encode("idna")
+        # When we have hostname we use it instead of netloc directly
+        if hostname:
+            if isinstance(username, str) and isinstance(password, str):
+                netloc_bytes += quote(username, _safe_chars).encode(encoding)
```
Gallaecio, line 112: Do we need `quote` here? I’m mostly concerned because `_safe_chars` contains some characters that may not be safe here, such as `:` for a username.

tests/test_url.py
Laerte: Done.

w3lib/url.py
Laerte: I have the same question, i follow the SO answer, what do you think? Or maybe we could remove the `_safe_chars` param.

Laerte: @Gallaecio Do you think worth add test for empty username or this is a edgy case?

Gallaecio:
I need to delay my review until I have the time to carefully look into the potential issues with encoding and decoding each separate field.

Gallaecio:
@Laerte Sorry for the wait, I finally had a proper look.

I have created a pull request against your branch, https://github.com/Laerte/w3lib/pull/1, please have a look when you get a chance.

`quote` was indeed needed for username and password, but `_safe_chars` was not the right character allow-list for those fields.

Laerte:
@Gallaecio Thank you! Do you think we can merge this to `master` or need more changes?

Gallaecio (approved):
It looks good to me. I’ll try and get someone else to have a look as well before we merge.

Laerte:
@Gallaecio Any update on this one?

wRAR (approved):

PR #185, opened 2022-08-08
Changelog

kmike:
I think we should release 2.0.0, because Python 2 compat is removed, and there are some deprecation removals.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,7 +24,7 @@ Deprecation removals (#169):
 The following functions are deprecated, and will be removed in future releases
 (#170):
 
-- `w3lib.util.str_to_unicode``
+- ``w3lib.util.str_to_unicode``
 - ``w3lib.util.unicode_to_str``
 - ``w3lib.util.to_native_str``
 
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,46 @@
 w3lib release notes
 ===================
 
+2.0.0 (to be released)
+----------------------
+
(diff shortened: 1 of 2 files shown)


NEWS
```diff
@@ -1,6 +1,46 @@
...
+The following functions are deprecated, and will be removed in future releases
+(#170):
+
+- `w3lib.util.str_to_unicode``
+- ``w3lib.util.unicode_to_str``
+- ``w3lib.util.to_native_str``
+
+Other improvements and bug fixes:
+
+- Type annotations are added (#172, #184).
```
kmike, line 33: #184 is not merged yet.

NEWS
```diff
@@ -1,6 +1,46 @@
...
+- ``w3lib.form`` module is removed.
+- ``w3lib.html.remove_entities`` function is removed.
+- ``w3lib.url.urljoin_rfc`` function is removed.
+
+The following functions are deprecated, and will be removed in future releases
+(#170):
+
+- `w3lib.util.str_to_unicode``
+- ``w3lib.util.unicode_to_str``
+- ``w3lib.util.to_native_str``
```
kmike, line 29: These functions are not documented as `:func:...`, because they're not in w3lib.rst auto-generated API docs, and I didn't want to add them there.

Gallaecio (approved):
I left suggestions at https://github.com/scrapy/w3lib/pull/186, but I am OK merging as is.

PR #190, opened 2022-09-23
update type annotation of auto_detect_fun param in html_to_unicode()

BurnzZ:


w3lib/encoding.py
```diff
@@ -218,7 +218,7 @@ def html_to_unicode(
     content_type_header: Optional[str],
     html_body_str: bytes,
     default_encoding: str = "utf8",
-    auto_detect_fun: Optional[Callable[[bytes], str]] = None,
+    auto_detect_fun: Optional[Callable[[bytes], Optional[str]]] = None,
```
BurnzZ, line 221: Fixes https://github.com/scrapinghub/web-poet/pull/81/files#diff-0e01d80b02cf2bc35a0ad4643b60cf8fd9f065721e497677d92e6af2fdf3ab89R290 since the `auto_detect_fun` callable could actually return a `None`.

kmike:
👍 nice catch @BurnzZ!

Issue #189, opened 2022-08-16
BOM should take precedence over Content-Type header when detecting the encoding


kmike:
Currently html_to_unicode prefers Content-Type header if BOM is present. 
But browsers, as well as WHATWG [standard](https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding) use BOM first. This can be checked e.g. by running this server, and opening URL in a browser - UTF-8 is used if BOM is present, and cp1251 is used if it's not:

```py
import codecs
from http.server import BaseHTTPRequestHandler
from http.server import HTTPServer


class HttpGetHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-type", "text/html; charset=cp1251")
        self.end_headers()
        self.wfile.write(codecs.BOM_UTF8)
        self.wfile.write("<!DOCTYPE html>".encode('utf8'))
        self.wfile.write("Привет!".encode('utf8'))


if __name__ == '__main__':
    httpd = HTTPServer(('', 8000), HttpGetHandler)
    httpd.serve_forever()

```

PR #191, opened 2022-09-23
update html_to_unicode() so that the BOM is used first to check the e…

BurnzZ:
Resolves https://github.com/scrapy/w3lib/issues/189

### DIFF
--- a/w3lib/encoding.py
+++ b/w3lib/encoding.py
@@ -227,8 +227,8 @@ def html_to_unicode(
 
     It will try in order:
 
-    * http content type header
     * BOM (byte-order mark)
+    * http content type header
     * meta or xml tag declarations
     * auto-detection, if the `auto_detect_fun` keyword argument is not ``None``
     * default encoding in keyword arg (which defaults to utf8)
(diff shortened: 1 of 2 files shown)


w3lib/encoding.py
```diff
@@ -281,27 +281,16 @@ def html_to_unicode(
...
-            html_body_str = html_body_str[len(bom) :]
-        elif enc == "utf-16" or enc == "utf-32":
-            # read endianness from BOM, or default to big endian
-            # tools.ietf.org/html/rfc2781 section 4.3
-            if bom_enc is not None and bom_enc.startswith(enc):
-                enc = bom_enc
-                bom = cast(bytes, bom)
-                html_body_str = html_body_str[len(bom) :]
-            else:
-                enc += "-be"
```
BurnzZ, line 300: I'm trying to add tests pertaining to this one but I'm not sure why this is needed.

```
>>> to_unicode(b'\xa3\x00', 'utf-16')
'£'
>>> to_unicode(b'\xa3\x00', 'utf-16-be')
'ꌀ'
>>> to_unicode(b'\xa3\x00', 'utf-16-le')
'�'
>>> to_unicode(b'\xff\xfe\x00\x00\xa3\x00\x00\x00', 'utf-32')
'£'
>>> to_unicode(b'\xff\xfe\x00\x00\xa3\x00\x00\x00', 'utf-32-be')
'��'
>>> to_unicode(b'\xff\xfe\x00\x00\xa3\x00\x00\x00', 'utf-32-le')
'\ufeff£'
```

Gallaecio: I get the same decoding for `utf-16` and `utf-16-le`:

```
>>> import w3lib
>>> w3lib.__version__
'2.0.1'
>>> from w3lib.encoding import to_unicode
>>> to_unicode(b'\xa3\x00', 'utf-16-le')
'£'
>>> to_unicode(b'\xa3\x00', 'utf-16')
'£'
```

BurnzZ: @Gallaecio what about the big endian value?

BurnzZ: That's weird, I now get the same value for both `utf-16` and `utf-16-le` as well. Perhaps I copy-pasted it wrong. 😅

Gallaecio: The output being the same for those 2 cases (but not for big endian), does your comment still apply?
The old code makes sense to me (i.e. apply big endian if the Content-Type matches the BOM, and the BOM indicates the endianess, which I take the Content-Type cannot or may not always indicate), and it makes sense to me that we do not need it anymore now that the BOM takes precedence.

BurnzZ: > it makes sense to me that we do not need it anymore now that the BOM takes precedence.

We still need it since the logic of adding "-be" as an encoding suffix only applies when the BOM is not present (or the bom is not the same with the content type headers).

In any case, I just realized that these tests in https://github.com/scrapy/w3lib/blob/master/tests/test_encoding.py#L254-L256 would fail if we do remove such logic. So I guess somehow it's still needed if the BOM isn't there.

wRAR (approved):

PR #197, opened 2022-10-22
Use latest version of Ubuntu

Laerte:
All workflows of Scrapy org runs on [`ubuntu-latest`](https://github.com/search?o=desc&q=org%3Ascrapy+runs-on&s=indexed&type=Code) runner, except for w3lib and [tests](https://github.com/scrapy/scrapyd/blob/master/.github/workflows/tests.yml) workflow of [scrapyd](https://github.com/scrapy/scrapyd) that's is being handled [here](https://github.com/scrapy/scrapyd/pull/457)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         include:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   publish:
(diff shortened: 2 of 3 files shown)


Gallaecio:
[I’m slightly against this change](https://github.com/scrapy/scrapy/issues/5671#issuecomment-1275873925).

Laerte:
@Gallaecio While i understand your point, i think the use case that you mentioned in the comment is more valid if the project rely on external packages/libraries.

In our case we don't rely in any external package/library. So if the python interpreter is successfully installed we should not have any failing behaviour changing our runner.

Gallaecio (approved):
That’s true. This is a pure Python package without dependencies. Which Ubuntu version we run should have no effect on tests. Thank you!

wRAR (approved):

PR #198, opened 2022-10-28
unit test Issue 91 is fixed

### LINKED ISSUE #91: `canonicalize_url` use of `safe_url_string` breaks when an encoded hash character is encountered

felipeboffnunes:
Resolves #91

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -845,12 +845,12 @@ class CanonicalizeUrlTest(unittest.TestCase):
             "http://www.example.com/a%A3do?q=r%E9sum%E9",
         )
 
-        url = "https://example.com/path/to/foo%20bar%3a%20biz%20%2376%2c%20bang%202017#bash"
+        url = "https://example.com/a%23b%2cc#bash"
         canonical = canonicalize_url(url)
         # %23 is not accidentally interpreted as a URL fragment separator
         self.assertEqual(
             canonical,
-            "https://example.com/path/to/foo%20bar:%20biz%20%2376,%20bang%202017"
+            "https://example.com/a%23b,c"
         )
         self.assertEqual(
             canonical,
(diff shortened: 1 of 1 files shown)


tests/test_url.py
```diff
@@ -820,6 +820,20 @@ def test_canonicalize_url_unicode_query_string_wrong_encoding(self):
...
+
+        # assert #bash is not on the url
+        canonical_url = (
+            "https://example.com/path/to/foo%20bar:%20biz%20%2376,%20bang%202017"
+        )
+        self.assertEqual(canonical, canonical_url)
+
+        # assert comma does not break url
+        canonical_from_canonical = canonicalize_url(canonical)
+        self.assertEqual(canonical, canonical_from_canonical)
```
Gallaecio, line 835: Maybe extend `test_normalize_percent_encoding_in_paths` instead, explaining that the test is focused on making sure that `%23` is not accidentally interpreted as a URL fragment separator.

felipeboffnunes: For the sake of simplicity, I could reduce the size of the URL as well

Gallaecio (approved):

kmike:
Thanks @felipeboffnunes!

Issue #193, opened 2022-10-09
Error when requesting ipv6 address directly with JOBDIR parameter


Cash111:

wRAR:
```
In [5]: safe_url_string('https://[2402:4e00:40:40::2:3b6]')
Out[5]: 'https://2402:4e00:40:40::2:3b6'

In [6]: safe_url_string('https://[2402:4e00:40:40::2:3b6]:80')
Out[6]: 'https://2402:4e00:40:40::2:3b6:80'
```

This indeed looks like a bug.

wRAR:
urlsplit returns `'[2402:4e00:40:40::2:3b6]:80'` in `netloc` but `2402:4e00:40:40::2:3b6` in `hostname`, and safe_url_string uses this hostname value directly without putting it in brackets again. There may be some code in urllib that should be used here instead.

himanshu007-creator:
Hi, i would like to work on this issue

Issue #80, opened 2016-12-12
Pipe symbol ("|") is not percent encoded


odinplus:
Pipe symbol ("|") is in reserved symbols list in url.py [https://github.com/scrapy/w3lib/blob/master/w3lib/url.py#L67](https://github.com/scrapy/w3lib/blob/master/w3lib/url.py#L67) and is not percent encoded by safe_url_string which used by scrapy to download urls. 
RFC mentioned in url.py [https://www.ietf.org/rfc/rfc3986.txt](https://www.ietf.org/rfc/rfc3986.txt) doesn't contain "|" in reserved symbols: 

```
      reserved    = gen-delims / sub-delims

      gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"

      sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

```
And I've found a site (in top 20 of Alexa, possible using play framework) which has such links with pipes, and it is answering with http code 400 (bad request) if "|" is not percent encoded in url. 
Is this a bug? How can I avoid it properly? For now I just removed "|" from url.py itself.

kmike:
There is a stalled PR to address that: https://github.com/scrapy/w3lib/pull/25

odinplus:
Yes, that PR addressing exactly this issue. Maybe if it is impossible to make general solution for safe symbols then it is worth to give optional control to the user? With some additional parameter in request.meta for example.

kmike:
@odinplus I wonder how this site works with Firefox, as according to @redapple's test Firefox doesn't encode `|` as well.

odinplus:
With Firefox it is answering with same code 400 if there is a pipe symbol in url.

kmike:
It seems there is still no consensus between browsers how to handle different characters in URL path (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1064700). This means that a website which works in one browser may break in another, and we can't create escaping method which works everywhere. #25 is merged, but we've removed `|` handling from it, so #25 does not fix this particular issue. `|` handling was removed for these reasons:

1) it makes changes smaller, more focused and less controversial;
2) Firefox handles `|` the same way as w3lib, so it is not that `|` handling is incorrect per se;
3) Chrome handles `|` differently in path and in query, while `safe_url_string` doesn't make this distinction currently, using the same set of chars - likely it should though.

[...]

PR #201, opened 2022-11-08
Make safe_url_string safer

Gallaecio:
Tasks:
- [x] Provide an alternative implementation for safe_url_string based on the URL living standard
- [x] Add missing typing information
- [x] Provide comprehensive test coverage for both implementations, highlighting the differences, and in the process make the new implementation support also RFC 2396 (used by [java.net.URI](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html)) and RFC 3986 (the previous standard, still popular).
- [x] Ensure the following issues are addressed by the new implementation:
    - [x] https://github.com/scrapy/w3lib/issues/80
    - [x] https://github.com/scrapy/w3lib/issues/193
- [x] Have the URL living standard parse implementation *mostly* pass the [upstream tests](https://github.com/web-platform-tests/wpt/blob/master/url/README.md).
- [ ] Improve performance
  The new implementation is currently up to 40 times slower than the previous implementation for test cases where both have the same, non-error outcome.
- [ ] Clean up the implementation: cleaner code, docstrings…
- [ ] Discuss what to do API-wise (deprecate safe_url_string in favor of safe_url? Remove safe_url and make safe_url_string use the new implementation?


Gallaecio:
> The new implementation is currently up to 40 times slower than the previous implementation for test cases where both have the same, non-error outcome.

@kmike Shall I look into using Cython here?

Gallaecio:
I have opened https://github.com/scrapy/w3lib/pull/203 as a simpler approach that is good enough to address #80 and similar issues without any major rework of `safe_url_string`.

I am closing this pull request, as I am stopping work on it. But I have created https://github.com/scrapy/w3lib/issues/204 so that someone, not necessarily me, can take over and finish the work.

PR #202, opened 2022-11-09
Handle `OverflowError` exception on  `convert_entity`

### LINKED ISSUE #199: Function `convert_entity` does not catch `OverflowError`
Error:
```
OverflowError
Python int too large to convert to C int
```

`w3lib` version -> `2.0.1`
`Python` version -> `3.8.13`

Screenshot 2022-10-31 at 10 02 22

Laerte:
Closes #199

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -66,6 +66,7 @@ class RemoveEntitiesTest(unittest.TestCase):
         self.assertEqual(replace_entities("x&#157;y"), "xy")
         self.assertEqual(replace_entities("x&#157;y", remove_illegal=False), "x&#157;y")
         self.assertEqual(replace_entities("&#82179209091;"), "")
+        self.assertEqual(replace_entities("&#82179209091;", remove_illegal=False), "&#82179209091;")
 
     def test_browser_hack(self):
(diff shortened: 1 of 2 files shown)


tests/test_html.py
```diff
@@ -65,6 +65,7 @@ def test_illegal_entities(self):
         self.assertEqual(replace_entities("x&#x2264;y"), "x\u2264y")
         self.assertEqual(replace_entities("x&#157;y"), "xy")
         self.assertEqual(replace_entities("x&#157;y", remove_illegal=False), "x&#157;y")
+        self.assertEqual(replace_entities("&#82179209091;"), "")
```
Gallaecio, line 68: I believe your fix covers it, but for the sake of test comprehension, I think it would be nice to cover an additional scenario:

```suggestion
        self.assertEqual(replace_entities("&#82179209091;"), "")
        self.assertEqual(replace_entities("&#82179209091;", remove_illegal=False), "&#82179209091;")
```

tomaszziebakalepa (approved):
Thanks for doing that! :)

wRAR (approved):

PR #205, opened 2022-11-27
Add release notes for version 2.1.0

Gallaecio:


wRAR (approved):

Gallaecio:
Let’s get https://github.com/scrapy/w3lib/pull/206 into the release.

wRAR:
Done!

NEWS
```diff
@@ -1,6 +1,63 @@
...
+    (#80, #203)
+
+-   :func:`~w3lib.encoding.html_to_unicode` now checks for the `byte order
+    mark`_ before inspecting the ``Content-Type`` header when determining the
+    content encoding, in line with the `URL living standard`_. (#189, #191)
+
+    .. _byte order mark: https://en.wikipedia.org/wiki/Byte_order_mark
+
+-   :func:`~w3lib.url.canonicalize_url` now strips spaces from the input URL,
+    to be more in line with the `URL living standard`_. (#132, #136)
```
kmike, line 38: hey @Gallaecio @wRAR! I think this is a wrong change without other changes, and we should either roll back, or change all other functions. See https://github.com/scrapy/w3lib/pull/136#issuecomment-1303746995.

kmike: Maybe! Actually, I'm not even sure my original comment is correct :) It may be a false alarm as well. `safe_url_string` docstring contains this:

> ASCII tabs and newlines are removed as per https://url.spec.whatwg.org/#url-parsing.

And indeed, there is a regex to remove tabs and newlines.

The issue still may stand that the whitespace handling is not compatible between these functions, but I'm not sure anymore.

We also would need to decide which functions in w3lib.url expect an already stripped URL (and may fail to work otherwise), and which can support more "raw" urls, with extra whitespace.

Gallaecio: https://github.com/scrapy/w3lib/pull/207 is ready for when we want to address this.

The Python docs say that, without arguments, [str.strip](https://docs.python.org/3/library/stdtypes.html#str.strip) removes "whitespace". From local tests, that seems to include more than the 3 characters `safe_url_string` escapes (it also removes e.g. `\f`), and I suspect it even includes non-ASCII space. If so, the change in `canonicalize_url` indeed goes too far, stripping characters that should not be stripped according to the URL living standard.

And I think both `canonicalize_url` and `safe_url_string` need to follow the living standard for URL parsing, which is the one that indicates the rules for stripping and removing characters from URLs before parsing. `canonicalize_url` should probably do the same as `safe_url_string` as a base, and then go further (e.g. query parameter sorting).

PR #206, opened 2022-11-27
Full typing

wRAR:


tests/test_url.py
```diff
@@ -315,24 +317,32 @@
 )
 
 
-def _test_safe_url_func(url, *, encoding=UNSET, output, func):
+def _test_safe_url_func(
+    url: StrOrBytes,
+    *,
+    encoding: Any = UNSET,
```
wRAR, line 323: `encoding` is actually either `str` or literally `UNSET`, but `UNSET = object()` and `typing.Literal` doesn't support literals of arbitrary types. It may be possible to refactor the code to make typing more simple.

tests/test_url.py
```diff
@@ -315,24 +317,32 @@
...
+    output: Union[str, Type[Exception]],
+    func: Callable[..., str],
+) -> None:
     kwargs = {}
     if encoding is not UNSET:
         kwargs["encoding"] = encoding
     try:
-        is_exception = issubclass(output, Exception)
```
wRAR, line 323: I feel like this try-except could be replaced with actual checks.

tests/test_url.py
wRAR: Moving this to the code.

Gallaecio: Using a sentinel value on a test file was probably overkill on my part, we can probably switch to `Optional[str]` and use `None` as sentinel.

Gallaecio (approved):

Issue #204, opened 2022-11-23
Reimplement safe_url_string based on the URL living standard


Gallaecio:
https://github.com/scrapy/w3lib/pull/201 implemented an alternative to `safe_url_string` based on the URL living standard, while maintaining support for older standards still used in the wild, for maximum safety of the resulting URL.

However, when we compared the performance of the new implementation with the old one, the new implementation was up to 200 times slower, and after some effort I only managed to reduce the gap to 40. 40 times slower is still unacceptable.

This task is about resuming the work started on #201, closing the performance gap by achieving a performance very similar or better than that of the current `safe_url_string` implementation.

How to do that? At the moment we believe [Cython](https://cython.readthedocs.io/en/latest/) is the way to go, i.e. rework the new implementation as Cython code, which hopefully will solve the performance issues.

wRAR:
Two points about Cython: first, I wonder if it's guaranteed to give comparable performance or is just an idea to try? Second, this will make w3lib a binary module, though both parsel and Scrapy already require lxml which is even more complicated.

kmike:
I think it might be reasonable to move the URL-related implementation to a separate package, to simplify contribution to the rest of w3lib - to work on w3lib itself it won't be required to have Cython installed. URL parsing according to Live Standard is a very well-defined scope, I think it could makes sense to have it separate even if it's pure Python.

It'd still be a binary dependency, but wheels work well these days, and unlike lxml there are no complications like the libxml dependency (though we might think about wrapping some external library, I haven't looked into it).

lopuhin:
https://github.com/mypyc/mypyc could be another option besides Cython, it may allow to keep the code python-compatible (but a separate package and a binary wheel would still be worthwhile).

PR #207, opened 2022-11-28
safe_url_string, canonicalize_url: apply stripping from the URL living standard

Gallaecio:


kmike (approved):
Looks good, thanks @Gallaecio!

Would it make sense to update safe_url_string docstring? It's no longer only tabs and newlines which are removed.

w3lib/url.py
```diff
@@ -77,9 +77,18 @@ def _quote_byte(error: UnicodeError) -> Tuple[str, int]:
 _FRAGMENT_SAFEST_CHARS = _PATH_SAFEST_CHARS
 
 
-_ascii_tab_newline_re = re.compile(
-    r"[\t\n\r]"
-)  # see https://infra.spec.whatwg.org/#ascii-tab-or-newline
+_ASCII_TAB_OR_NEWLINE = "\t\n\r"
```
wRAR, line 80: Doesn't this duplicate w3lib._infra?

Gallaecio:
Is https://github.com/scrapy/w3lib/pull/207/commits/288caae88640e0243aed1a7f8e73047c639e324a OK? It describes more what the functions should do than what they actually do.

wRAR (approved):

w3lib/url.py
```diff
@@ -531,11 +556,8 @@ def canonicalize_url(
 ) -> str:
     r"""Canonicalize the given url by applying the following procedures:
 
+    - make the URL safe (see :func:`safe_url_string`)
```
kmike, line 559: That's actually a question I had :) It seems canonicalize_url doesn't call safe_url_string. How do we ensure it makes the URl safe in the same way? Or is it not a requirement? From the docstring I'd infer that canonicalize_url calls safe_url_string, and then does something on top of it.

w3lib/url.py
```diff
@@ -88,9 +95,29 @@ def safe_url_string(  # pylint: disable=too-many-locals
...
+    """Return a URL equivalent to *url* that a wide range of web browsers and
+    web servers consider valid.
+
+    *url* is parsed according to the rules of the `URL living standard`_,
+    and during serialization additional characters are percent-encoded to make
+    the URL valid by additional URL standards.
+
+    .. _URL living standard: https://url.spec.whatwg.org/
+
+    The returned URL is valid by *all* of the following URL standards known to
```
kmike, line 107: @Gallaecio I think you mentioned that the implementation is not 100% compatible, until we solve https://github.com/scrapy/w3lib/issues/204. Is there some detail to it?

Gallaecio: No, there is not. This is why I said:

> It describes more what the functions should (in my opinion) do than what they actually do.

Any suggestion on how to handle this? Should we say that the function *aims* to do this, but there are know issues or divergences at the moment?

kmike: A small language tweak ("should be valid" instead of "is valid") might be enough.

w3lib/url.py
kmike: If we revert, it won't be worse than before, but the docstring will not be correct - it mentions many things, but not the whitespace handlng. So, some options:

1. Revert the docstring change
2. Write "make the URL safe" without referring to safe_url_string
3. Document that whitespace and control characters are stripped

I'm fine with all 3 :)

kmike:
Thanks @Gallaecio!

PR #233, opened 2025-01-21
test_url.py: handle 3.9 and 3.10 for http://[2a01:5cc0:1:2:3:4] testcase

flokli:
The fix for https://github.com/python/cpython/issues/103848 / CVE-2024-11168 also got backported to 3.9.21 and 3.10.16, so we cannot add that testcase for these versions either.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -420,7 +420,11 @@ KNOWN_SAFE_URL_STRING_URL_ISSUES = {
     # (%) are not escaped.
     f"a://example.com#{FRAGMENT_TO_ENCODE}",
 }
-if sys.version_info < (3, 11, 4):
+if (
+    sys.version_info < (3, 9, 21)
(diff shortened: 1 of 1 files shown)


tests/test_url.py
```diff
@@ -420,7 +420,7 @@ def test_safe_url_string_encoding(
     # (%) are not escaped.
     f"a://example.com#{FRAGMENT_TO_ENCODE}",
 }
-if sys.version_info < (3, 11, 4):
+if not(sys.version_info >= (3, 9, 21) or sys.version_info >= (3, 10, 16) or sys.version_info >= (3, 11, 4)):
```
wRAR, line 423: I don't think it works this way? 3.10.16 is already >= 3.9.21 (as is 3.10.0).

flokli: ah right, we need to conditionalize on 3.9/3.10 too. Will update.

Gallaecio (approved):

wRAR (approved):
Thanks!

PR #237, opened 2025-01-30
Migrate to ruff.

wRAR:
This also kills the trivial `w3lib._types.StrOrBytes` alias.


.coveragerc
```diff
@@ -1,3 +1,7 @@
 [run]
 branch = true
 include = w3lib/*
+
+[report]
+exclude_lines =
+    if TYPE_CHECKING:
```
Gallaecio, line 7: This may be a good chance to move this config to `pyproject.toml`, [it seems to be supported](https://coverage.readthedocs.io/en/latest/config.html).

Gallaecio (approved):

PR #247, opened 2025-07-27
tests: benchmarks, xdist

albertedwardson:
This pull request introduces: 
- a CodSpeed-powered benchmarking workflow with addition of benchmark test suites 
- parallelization of existing tests using pytest-xdist


albertedwardson:
@wRAR @AdrianAtZyte hi! what do you think about enabling codspeed integration? ci seems to work, it fails at report level, as it's not set up in this repo/org https://codspeed.io/docs/integrations/providers/github 

img

albertedwardson:
note to reviewers: the `BENCHMARK_CASES` dict is generated with help of AI, no need to validate it diligently, until you will be ok with enabling codspeed in ci. after that- I will do a self-review, disable supression of exceptions (and maybe do asserts with expected output, idk)

albertedwardson:
I should definitely group all test by function, rather than parameterize them, so they don't reported as isolated cases

since, imo, an overall performance report would be more informative than fine-grained report for each edge case

albertedwardson:
I'm having problems with reports even on my fork: https://github.com/albertedwardson/w3lib/pull/1

Even though the job logs indicate that the results have been uploaded, they aren't showing up in the codspeed ​​web ui

Perhaps due to an incorrect cache key (that is somehow set by concurrency param in workflow file)? idk

https://github.com/albertedwardson/w3lib/actions/runs/24640475739/job/72043441687#step:6:1

AdrianAtZyte:
I did notice the upstream docs mention the need for `id-token: write` permission in actions.

That said, I *suspect* the main reason https://codspeed.io/scrapy/w3lib/branches/albertedwardson%3Apytest-benchmark does not show much is because until the benchmarks run on the main branch there is no baseline to compare with.

albertedwardson:
`id-token: write` wasn't the cause; the logs shows `NOTICE: codspeed is enabled, but no performance measurement will be made since it's running in an unknown environment`, although codspeed should definitely know that it's in CI

maybe it doesn't like being run through `tox`?

albertedwardson:
yeah, `tox` was the cause... I'll try find a way to convince codspeed ​​that running with `tox` is ok, it only supports `nox`...

albertedwardson:
ok, pypy is not supported: https://github.com/CodSpeedHQ/pytest-codspeed/issues/34

albertedwardson:
I thought everything was working locally because I didn't notice the `notice` log :)

```
NOTICE: codspeed is enabled, but performance measurements won't be performed because it's running in an unknown environment.
```

This means it doesn't even attempt to run the instrumentation, but only checks whether the tests pass

albertedwardson:
idk how to do this nicely without excessive copypaste from real test cases

pyproject.toml
```diff
@@ -78,6 +78,7 @@ strict = true
 # Allow test functions to be untyped
 module = "tests.*"
 allow_untyped_defs = true
+disallow_untyped_calls = false
```
AdrianAtZyte: Can we set this only for `tests.benchmarks.*`?

tox.ini
```diff
@@ -4,16 +4,17 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py310, py311, py312, py313, py314, pypy3.11, docs, pylint, typing, pre-commit, twinecheck
+envlist = py310, py311, py312, py313, py314, pypy3.11, docs, pylint, typing, pre-commit, twinecheck, benchmark
```
AdrianAtZyte, line 7: No strong opinion, but I wonder if running it by default with `tox` makes sense. Do we want people to run them before pushing changes? Do they tell you the performance difference between your changes and the code in the main branch? I assume these are more useful in Codspeed, where I assume it tracks that difference and warns on significant degradation.

albertedwardson: running codspeed like that locally will not produce any reports, it will only confirm that benchmarks can be run, so benchmarks are treated as normal tests (with assertion errors and exceptions)

AdrianAtZyte (approved):

PR #248, opened 2025-07-27
fix wrongly running macos workflow on ubuntu-latest + formatting

abebus:


.github/workflows/tests-macos.yml
```diff
@@ -10,26 +10,37 @@ on:
 
 jobs:
   tests:
+    runs-on: ${{ matrix.os }}
 
-    runs-on: ubuntu-latest
```
wRAR, line 14: :facepalm:

.github/workflows/tests-macos.yml
```diff
@@ -10,26 +10,37 @@ on:
...
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14.0-rc.1"]
+        os:
+          - macos-latest
+          - macos-13
```
wRAR, line 20: No, it isn't needed, we only use `macos-latest` everywhere.

wRAR (approved):

Gallaecio (approved):
I don’t love the formatting, but I am OK with it as long as we extend the pre-commit config to apply it automatically to every YAML file in the repo, e.g. with https://github.com/google/yamlfmt.

abebus:
@wRAR @Gallaecio should I set space offset to 2 instead of the defaults 4?

wRAR:
I don't have preferences for a yaml formatting style so anything works for me :)

PR #256, opened 2026-04-27
Improve performance and readability of some `w3lib.html` functions

albertedwardson:


AdrianAtZyte:
The coverage data shows that `if offset < start:` and `if offset < len(utext):` are always true with the current tests. Is that because they cannot be false (and hence the ifs are not needed), or can we add one or two tests to cover the scenarios where they evaluate to false?

AdrianAtZyte (approved):

wRAR (approved):

wRAR:
Thanks!

PR #257, opened 2026-05-02
Improve performance of `w3lib.url` functions

albertedwardson:
This PR includes:
- Tool version updates: mypy 1.19.1 -> 1.20.2 (although mypy 2.0 has already been released), ruff 0.14.14 -> 0.15.12, pylint 4.0.4 -> 4.0.5
- Custom and faster implementations of some `urllib.parse` functions, which remain private
- Added early exits that skip resource-intensive work and cache it where necessary
- Backport of `_check_bracketed_netloc` from https://github.com/python/cpython/issues/105704
- Minor code refactoring in `w3lib/url.py`

The most significant improvement is seen in `safe_url_string`, as its speed has increased by 4.2x.

note: this PR does not use Cython, mypyc, or performance-oriented URL parsing libraries, and it does not include any new dependencies that might provide even better performance


albertedwardson:
original `pathname2url` somehow utilities SIMD, idk how

tests/test_url.py
```diff
@@ -1440,14 +1561,14 @@ def test_canonicalize_url_idempotence(self):
...
+            == "http://.éxamplé.com/r%C3%A9sum%C3%A9?q=r%C3%A9sum%C3%A9"
         )
 
         # DNS label too long
         assert (
-            canonicalize_url(f"http://www.{'example' * 11}.com/résumé?q=résumé")
-            == f"http://www.{'example' * 11}.com/r%C3%A9sum%C3%A9?q=r%C3%A9sum%C3%A9"
+            canonicalize_url(f"http://www.{'éxamplé' * 11}.com/résumé?q=résumé")
+            == f"http://www.{'éxamplé' * 11}.com/r%C3%A9sum%C3%A9?q=r%C3%A9sum%C3%A9"
         )
```
albertedwardson, line 1572: I've changed hostnames so they trigger idna encoding, my new helper function skips it if input is fully ascii

w3lib/_url.py
```diff
@@ -10,3 +34,763 @@
...
+    path: str
+    query: str
+    fragment: str
+
+    username: str | None = None
+    password: str | None = None
+    hostname: str | None = None
+    port: str | int | None = None
+
+    def __post_init__(self) -> None:
```
albertedwardson, line 511: this class is definitely not a drop in replacement for std urlib SplitResult, but it is used for convenience and it's private anyway 

the main behavioural change is that it raises error about invalid port at initialisation level, when std SplitResult does this when port property is accessed

w3lib/encoding.py
```diff
@@ -140,7 +140,7 @@ def _c18n_encoding(encoding: str) -> str:
     encoding aliases
     """
     normed = encodings.normalize_encoding(encoding).lower()
-    return cast("str", encodings.aliases.aliases.get(normed, normed))
```
albertedwardson, line 143: it was a bug in typeshed, which I fixed btw :)

w3lib/url.py
```diff
@@ -209,18 +232,18 @@ def safe_download_url(
...
+            path = f"{path}/"
     else:
         path = "/"
-    return urlunsplit((scheme, netloc, path, query, ""))
+    return _urlunsplit((scheme, netloc, path, query, ""))
 
 
 def is_url(text: str) -> bool:
-    return text.partition("://")[0] in ("file", "http", "https")
+    return text.partition("://")[0] in {"file", "http", "https"}
```
albertedwardson, line 231: it's a ruff suggestion

w3lib/_url.py
```diff
@@ -10,3 +34,763 @@
...
+    lower-level optimisation involved is Unicode normalization
+    (NFKC), which may use optimized internal paths, but IDNA itself
+    remains Python-level logic.
+    """
+    if input_string.isascii():
+        return input_string.encode(), input_string
+
+    _, normalized = _nfkc_netloc(input_string)
+
+    encoded = normalized.encode("idna")
```
albertedwardson, line 769: cpython has fast check if string is already normalised, so this is not wasteful

w3lib/url.py
```diff
@@ -10,55 +10,61 @@
...
+from ._infra import (
+    _ASCII_TAB_OR_NEWLINE as _ASCII_TAB_OR_NEWLINE,
+    _C0_CONTROL_OR_SPACE as _C0_CONTROL_OR_SPACE,
+)
+from ._url import (
+    # reexports
+    _ASCII_TAB_OR_NEWLINE_TRANSLATION_TABLE as _ASCII_TAB_OR_NEWLINE_TRANSLATION_TABLE,
+    _PATH_SAFE_CHARS as _path_safe_chars,
+    _SAFE_CHARS as _safe_chars,
+    _SPECIAL_SCHEMES as _SPECIAL_SCHEMES,
```
albertedwardson: maybe dont reexport those since they're private?

albertedwardson:
@codspeedbot explain why this is faster

w3lib/url.py
AdrianAtZyte: Yes, no need to re-export anything private.

pyproject.toml
```diff
@@ -233,10 +238,17 @@ ignore = [
...
+    "PLC0414",
+    # Too many positional arguments
+    "PLR0917",
+    # Private name import
+    "PLC2701",
 ]
 
 [tool.ruff.lint.isort]
 split-on-trailing-comma = false
+combine-as-imports = true
```
AdrianAtZyte, line 248: 👍

pyproject.toml
```diff
@@ -233,10 +238,17 @@ ignore = [
     "RUF012",
     # Use of `assert` detected
     "S101",
+    # Import alias does not rename original package
+    "PLC0414",
```
AdrianAtZyte, line 239: Why ignore this one? https://docs.astral.sh/ruff/rules/useless-import-alias/#useless-import-alias-plc0414 makes sense, is it misreporting something?

pyproject.toml
```diff
@@ -107,8 +107,13 @@ disable = [
     "unused-argument",
     "unused-import",
     "unused-variable",
+    "useless-import-alias",
+    "possibly-used-before-assignment",
 ]
 
+[tool.ruff]
+target-version = "py310"
```
AdrianAtZyte: I believe `requires-python = ">=3.10"` above makes this unnecessary.

AdrianAtZyte (commented):
This is amazing, great work!

pyproject.toml
albertedwardson: This aliasing is intentional for explicit re-exports:

```python
    RFC3986_GEN_DELIMS as RFC3986_GEN_DELIMS,
    RFC3986_RESERVED as RFC3986_RESERVED,
    RFC3986_SUB_DELIMS as RFC3986_SUB_DELIMS,
    RFC3986_UNRESERVED as RFC3986_UNRESERVED,
    RFC3986_USERINFO_SAFE_CHARS as RFC3986_USERINFO_SAFE_CHARS,
```

 `PLC0414` is a known false positive in this pattern. Ruff itself documents redundant aliases as the supported way to mark re-exports for `F401`/public API handling.

`__all__` only defines `from module import *` behavior; it’s not the same thing as explicit re-export declarations.

AdrianAtZyte: We do usually use all to silence these cases, and I think most IDEs will warn you when importing something not in __all__. That said, your explanation makes sense, no need to switch to __all__.

AdrianAtZyte (approved):

wRAR (approved):
Thanks!

PR #259, opened 2026-05-31
Property-based tests for w3lib.url.

wRAR:
This is my first experience with `hypothesis` and these tests are pretty basic but by coincidence this may be the best time to add them after recent refactoring but before the release. Note that some of these fail, I'll let @albertedwardson decide what to skip/omit and what to fix.


albertedwardson:
I'll take a look at that in a week or two

wRAR:
@AdrianAtZyte asked if these tests pass or fail on 2.4.1. The only test outside of `TestPrivateHelpersProperties` that fails on this PR is `TestCanonicalizeUrlProperties::test_idempotent`, it passes on 2.4.1.

albertedwardson:
Sorry guys, I’ll be available only next month. I’ve had a lot of other things to do lately

wRAR:
:-/

tests/test_url.py
```diff
@@ -1632,6 +1720,21 @@ def test_strip_spaces(self):
         assert canonicalize_url(" https://example.com ") == "https://example.com/"
 
 
+class TestCanonicalizeUrlProperties:
+    @given(hyp_urls())
+    def test_no_exception(self, url: str) -> None:
+        canonicalize_url(url)
+
+    @given(st.text() | hyp_urls())
+    @example(r"\?").xfail()
```
albertedwardson: This idempotency test will be fixed by https://github.com/scrapy/w3lib/commit/577c9b530de2b75462056cd7c0b4208113f66b83

It fails because on the second pass `_urlsplit` assumed one single square bracket in path as a malformed ipv6

Also, I would exclude all non-url inputs from tests, since those private functions expect a valid url as an input 

```suggestion
    @given(hyp_urls())
```

tests/test_url.py
```diff
@@ -1800,3 +1903,78 @@ def test_urlunsplit(self, components, expected):
     )
     def test_urlunparse(self, components, expected):
         assert _urlunparse(*components) == expected
+
+
+class TestPrivateHelpersProperties:
+    @example("/").xfail()
+    @given(st.text())
+    def test_quote_matches_stdlib(self, data: str) -> None:
+        result = _quote(data.encode("utf-8")).decode("utf-8")
```
albertedwardson:
```suggestion
    @given(st.text())
    def test_quote_matches_stdlib(self, data: str) -> None:
        result = _quote(data.encode("utf-8"), safe=b"/").decode("utf-8")
```

tests/test_url.py
```diff
@@ -1800,3 +1903,78 @@ def test_urlunsplit(self, components, expected):
...
+        assert result == expected
+
+    @given(st.text())
+    def test_parse_qsl_matches_stdlib(self, data: str) -> None:
+        result = _parse_qsl(data)
+        expected = parse_qsl(data.encode("utf-8"))
+        assert result == expected
+
+    @given(st.text() | hyp_urls())
+    @example(":").xfail()
```
albertedwardson:
```suggestion
    @given(hyp_urls())
```

tests/test_url.py
```diff
@@ -1800,3 +1903,78 @@ def test_urlunsplit(self, components, expected):
...
+        result = _urlparse(url)
+        expected = urlparse(url)
+        assert result.scheme == expected.scheme
+        assert result.netloc == expected.netloc
+        assert result.path == expected.path
+        assert result.query == expected.query
+        assert result.fragment == expected.fragment
+
+    @given(st.text() | hyp_urls())
+    @example("0\n").xfail()
```
albertedwardson:
```suggestion
    @given(hyp_urls())
```

albertedwardson (commented):
with the latest fixes merged to master and my suggestions, the tests will be fixed (hopefully)

tests/test_url.py
albertedwardson: to fully match stdlib we need to pass `safe=b"/"` into it, since our implementation of `_quote` either called on a parts of url where slash should be quoted, or with explicitly passing `_PATH_SAFE_CHARS` as safe chars

tests/test_url.py
```diff
@@ -1226,6 +1250,70 @@ def test_any_to_uri(self):
...
+    def test_no_exception(self, url: str) -> None:
+        add_or_replace_parameters(url, {"foo": "bar"})
+
+
+class TestSafeUrlStringProperties:
+    @given(hyp_urls())
+    def test_no_exception(self, url: str) -> None:
+        safe_url_string(url)
+
+    @given(st.text() | hyp_urls())
```
albertedwardson: i think this also should only test on urls, not on random text

tests/test_url.py
```diff
@@ -1648,6 +1736,20 @@ def test_strip_spaces(self):
         assert canonicalize_url(" https://example.com ") == "https://example.com/"
 
 
+class TestCanonicalizeUrlProperties:
+    @given(hyp_urls())
+    def test_no_exception(self, url: str) -> None:
+        canonicalize_url(url)
+
+    @given(st.text() | hyp_urls())
```
albertedwardson: same here

tests/test_url.py
```diff
@@ -1226,6 +1250,70 @@ def test_any_to_uri(self):
...
+            return
+        assert safe_url_string(once) == once
+
+
+class TestParseUrlProperties:
+    @given(hyp_urls())
+    def test_no_exception(self, url: str) -> None:
+        parse_url(url)
+
+    @given(st.text() | hyp_urls())
```
albertedwardson: same here

tests/test_url.py
```diff
@@ -1825,3 +1927,76 @@ def test_urlunsplit(self, components, expected):
...
+    @given(st.text())
+    def test_quote_plus_matches_stdlib(self, data: str) -> None:
+        result = _quote(data.encode("utf-8"), quote_plus=True).decode("utf-8")
+        expected = quote_plus(data)
+        assert result == expected
+
+    @given(st.text())
+    def test_unquote_matches_stdlib(self, data: str) -> None:
+        result = _unquote(data)
+        expected = unquote(data).encode("utf-8")
```
albertedwardson: default safe differs here too

albertedwardson:
I think we should add guard for exceptions, but with a check that if the stdlib variant threw an exception, then our variant did too

wRAR:
This is finally passing.

AdrianAtZyte (approved):

PR #221, opened 2024-02-13
Implement a safe_url based on all standards

Gallaecio:
Continuation of https://github.com/scrapy/w3lib/pull/201.

Fixes https://github.com/scrapy/w3lib/issues/193.

Tasks:
- [x] Provide an alternative implementation for safe_url_string based on the URL living standard
- [x] Add missing typing information
- [x] Provide comprehensive test coverage for both implementations, highlighting the differences, and in the process make the new implementation support also RFC 2396 (used by [java.net.URI](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html)) and RFC 3986 (the previous standard, still popular).
- [x] Ensure the following issues are addressed by the new implementation:
    - [x] https://github.com/scrapy/w3lib/issues/80
    - [x] https://github.com/scrapy/w3lib/issues/193
- [x] Have the URL living standard parse implementation *mostly* pass the [upstream tests](https://github.com/web-platform-tests/wpt/blob/master/url/README.md).
- [ ] Improve performance
  The new implementation is currently around 3-4 times slower than the previous implementation for test cases where both have the same, non-error outcome.
- [ ] Address to-dos in `test_parse_url`.
- [ ] Clean up the implementation: cleaner code, docstrings…
- [ ] Discuss what to do API-wise (deprecate safe_url_string in favor of safe_url? Remove safe_url and make safe_url_string use the new implementation?

PR #261, opened 2026-06-10
Use can_ada for supported URLs

### LINKED ISSUE #204: Reimplement safe_url_string based on the URL living standard

### LINKED ISSUE #222: canonicalize_url incorrectly handles port when using hostname that requires IDNA encoding
Hello,

```python
```
which returns

while the expected value is

### LINKED ISSUE #1306: Speedup & fix URL parsing
I profiled a simple Scrapy spider which just downloads pages and follows links extracted using LinkExtractor; it turns out one of the main bottlenecks is urlparse module and our related functions like canonicalize_url or request_fingerprint. They take several times more CPU time than e.g. HTML parsing itself. 

Also, these functions don't follow modern HTML5 standards - see e.g. https://github.com/scrapy/scrapy/issues/1304

I wonder if we can fix it all in once by wrapping an actual browser implementation. Some options:

* https://code.google.com/p/google-url/
* http://mxr.mozilla.org/mozilla-central/source/netwerk/base/nsURLParsers.cpp
* http://src.chromium.org/viewvc/chrome/trunk/src/url/
* https://github.com/servo/rust-url

Modern tests for URL parsing: https://github.com/w3c/web-platform-tests/blob/master/url/urltestdata.txt

AdrianAtZyte:
https://github.com/scrapy/w3lib/pull/221#issuecomment-2002153108

Fixes https://github.com/scrapy/scrapy/issues/1306, fixes #98, fixes #204, closes #221, fixes #222, closes #270.

Some of the tests in #259 would still not pass with these changes, but I suspect that may be OK (can_ada / WHATWG diverge from stdlib).


tests/test_html.py
wRAR: I think this is changed with the expectation change

tests/test_url.py
```diff
@@ -660,7 +662,7 @@ def test_safe_url_string_bytes_input(self):
             b"http://www.example.com/\xa3?unit=\xb5", encoding="latin1"
         )
         assert isinstance(safeurl, str)
-        assert safeurl == "http://www.example.com/%C2%A3?unit=%B5"
+        assert safeurl == "http://www.example.com/%C2%A3?unit=%C2%B5"
```
wRAR, line 710: There are several changes that look like it was latin1 and now is utf-8, and in at least some of them an arg with `"latin1"` is passed to the function, I wonder if all of these are still valid behavior.

wRAR:
> Some of the tests in #259 would still not pass with these changes, but I suspect that may be OK (can_ada / WHATWG diverge from stdlib).

Yes, it says that explicitly so maybe we don't want to have those tests...

tests/test_url.py
AdrianAtZyte: My understanding is that latin1 is used for decoding bytes to str, then utf-8 for percent-encoding, which I think aligns with WHATWG.

albertedwardson:
I'd keep can_ada as an optional dependency and also used it's `idna_decode` and `idna_encode`

AdrianAtZyte:
Any specific reason in mind for making it optional? I’m not against it, but I would usually do that only if I can think of a good reason why someone would not want it installed (e.g. not supporting or being hard to install on some platform). I did not check if can_ada can be easily installed on Windows, and I did notice that the last commit was from 6 months ago which may be worrisome (although maybe we could assume maintenance in a worst-case scenario if it seems worth it, which it does to me). I wonder if you have other reasons in mind.

albertedwardson:
Before this PR, w3lib had no runtime dependencies at all. This would introduce the first one, and it’s also a compiled extension. Some users *may* be relying on w3lib specifically because it’s dependency-free and works everywhere without requiring a compiler or prebuilt wheels.

I’m not saying we shouldn’t use can_ada, just that changing those assumptions can have downstream effects on existing environments and build pipelines. To me, that’s either something for a major version bump, or a good case for making it an optional dependency so existing users aren’t affected unless they opt in.

PR #264, opened 2026-06-14
Avoid quadratic backtracking in get_base_url (fixes #263)

### LINKED ISSUE #263: w3lib.html.get_base_url() can be slow due to quadratic regex complexity
`get_base_url("<base "*30000)` takes several second here, due to `<base\s[^>]*` in `w3lib.html._baseurl_re`. Changing `[^>]` to `[^<>]` fixes this case, not sure if there are still problematic ones after that.

Note that Scrapy only passes the first 4Kb of the page to this function so it's not a problem there.

tienvantranspk-tech:
Fixes #263.

## Problem
`w3lib.html._baseurl_re` is `<base\s[^>]*href\s*=...`. The `[^>]*` matches everything up to the next `>`. On input with many `<base ` starts and no closing `>` (for example `<base ` repeated tens of thousands of times), the regex backtracks quadratically, so `get_base_url()` takes several seconds:

    n= 5000   0.36s
    n=10000   1.44s   (4x for 2x input -> O(n^2))
    n=20000   5.91s

(Scrapy itself only passes the first 4 KB to this function, so it is not affected there, but other callers can be.)

## Fix
Use `[^<>]` instead of `[^>]`. An HTML tag's content cannot contain a raw `<`, so this still matches every valid `<base ...>` tag, but stops the `*` from scanning across tag boundaries. The match becomes linear:

    n=10000   0.001s
    n=20000   0.001s
    n=40000   0.001s

## Tests
Added `TestGetBaseUrl.test_get_base_url_redos`, checking both that the pathological input is handled (returning the fallback base URL) and that a real `<base href>` after such a prefix is still found. Full suite passes (355 passed); `ruff check` / `ruff format` are clean.

Note: `get_meta_refresh()` shows similar quadratic behavior on analogous input, but via a different code path (not `_baseurl_re`). I kept this PR focused on #263 and am happy to look into that separately.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -363,11 +363,6 @@ class TestGetBaseUrl:
         )
 
     def test_get_base_url_no_catastrophic_backtracking(self):
-        # Regression test for https://github.com/scrapy/w3lib/issues/263:
-        # many "<base " starts with no closing ">" made _baseurl_re backtrack
-        # quadratically (tens of seconds for tens of thousands of starts). The
-        # time bound is deliberately generous: the fixed code runs in well under
-        # a millisecond, so a quadratic regression would blow past 2 seconds.
         prefix = "<base " * 30000
         start = time.perf_counter()
         assert get_base_url(prefix, "http://example.com/") == "http://example.com/"
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -1,3 +1,5 @@
+import time
+
 from w3lib.html import (
     get_base_url,
     get_meta_refresh,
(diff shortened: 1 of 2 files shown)


tests/test_html.py
```diff
@@ -360,6 +360,20 @@ def test_get_base_url(self):
             == "http://example.org/something"
         )
 
+    def test_get_base_url_redos(self):
+        # Regression test for https://github.com/scrapy/w3lib/issues/263:
+        # many "<base " starts with no closing ">" made _baseurl_re backtrack
+        # quadratically. Extraction must stay fast and still find a real tag.
```
wRAR: How does this test that "Extraction must stay fast"?

tienvantranspk-tech: Good catch - it didn't; it only asserted correctness, so a regression would have made the test slow rather than fail. I've pushed a commit that adds an explicit (generous) time bound, so a quadratic regression now fails outright: the fixed code runs in well under a millisecond here, while the previous regex took tens of seconds at this input size.

Happy to express this as a CodSpeed benchmark under `tests/benchmarks/` instead, if you'd prefer that to a wall-clock assertion.

AdrianAtZyte (approved):

wRAR:
Confirmed that the test fails without the change.

wRAR (approved):

PR #266, opened 2026-06-15
Avoid quadratic backtracking in get_meta_refresh (fixes #265)

### LINKED ISSUE #265: get_meta_refresh() is quadratic on malformed HTML (catastrophic backtracking)
`get_meta_refresh()` shows quadratic (catastrophic-backtracking) behaviour on malformed input, the same class of issue as #263 (which was about `get_base_url`).

### Reproducer

    import time
    from w3lib.html import get_meta_refresh

    for n in (5000, 10000, 20000):
        html = "<meta " * n
        t = time.perf_counter()
        get_meta_refresh(html)
        print(n, round(time.perf_counter() - t, 3))

Output:

    5000   0.069
    10000  0.266
    20000  1.529   # ~4x for 2x input -> O(n^2)

### Cause

`_meta_tag_re = re.compile(r"<meta\b[^>]*>", re.IGNORECASE)`. On input with many `<meta ` starts and no closing `>`, the `[^>]*` backtracks quadratically (the same pattern fixed for `_baseurl_re` in #263).

### Suggested fix

Use `[^<>]` instead of `[^>]` so the `*` cannot scan across tag boundaries; a tag's content cannot contain a raw `<`, so valid `<meta>` tags still match. PR to follow.

tienvantranspk-tech:
Fixes #265.

## Problem
`get_meta_refresh()` is quadratic on malformed input, the same class of bug as #263 (`get_base_url`). The cause is `_meta_tag_re`:

    _meta_tag_re = re.compile(r"<meta\b[^>]*>", re.IGNORECASE)

On input with many `<meta ` starts and no closing `>` (e.g. `"<meta " * n`), `[^>]*` backtracks quadratically:

    n= 5000   0.07s
    n=10000   0.27s   (~4x for 2x input -> O(n^2))
    n=20000   1.53s

## Fix
Use `[^<>]` instead of `[^>]`, exactly as in #263. An HTML tag's content cannot contain a raw `<`, so valid `<meta>` tags still match, while the `*` can no longer scan across tag boundaries. `get_meta_refresh` becomes linear (sub-millisecond at the sizes above).

## Tests
Added `TestGetMetaRefresh.test_get_meta_refresh_no_catastrophic_backtracking` with a generous time bound (so a quadratic regression fails the test) that also checks a real refresh `<meta>` after the pathological prefix is still parsed. Full suite passes (355 passed); `ruff check` / `ruff format` are clean.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -466,11 +466,6 @@ class TestGetMetaRefresh:
         assert get_meta_refresh(body, baseurl) == (5, "http://example.org/newpage")
 
     def test_get_meta_refresh_no_catastrophic_backtracking(self):
-        # Regression test for https://github.com/scrapy/w3lib/issues/265:
-        # many "<meta " starts with no closing ">" made _meta_tag_re backtrack
-        # quadratically (same class of bug as #263 for _baseurl_re). The time
-        # bound is deliberately generous; the fixed code runs in well under a
-        # millisecond, so a quadratic regression would blow past 2 seconds.
         prefix = "<meta " * 30000
         start = time.perf_counter()
         assert get_meta_refresh(prefix) == (None, None)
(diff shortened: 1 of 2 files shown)


AdrianAtZyte (approved):

wRAR:
The test passes locally without the change, as the threshold is too low for fast machines, so I've increased the input length.

wRAR (approved):

PR #271, opened 2026-06-19
avoid quadratic backtracking in remove_tags_with_content

uchiha-bug-hunter:
`remove_tags_with_content("<script " * n, which_ones=("script",))`:

    n=10000    0.56s
    n=20000    2.05s
    n=40000   10.68s

`_build_remove_tags_pattern` matches a tag's attribute span with `<(tag)\b[^>]*>`. On input with many `<script ` starts and no closing `>`, the `[^>]*` backtracks quadratically. Same case as #263 (`_baseurl_re`) and #265 (`_meta_tag_re`): a tag cannot contain a raw `<`, so `[^<>]` still matches valid tags while stopping the scan from crossing tag boundaries, making it linear.

`get_meta_refresh` reaches the same path, since it strips `script`/`noscript` via `remove_tags_with_content` by default.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -277,6 +277,18 @@ class TestRemoveTagsWithContent:
             == "<span></span>"
         )
 
+    def test_no_catastrophic_backtracking(self):
+        evil = "<script " * 50000
+        start = time.perf_counter()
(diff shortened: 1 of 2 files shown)


AdrianAtZyte (approved):
(verified the test fails without the fix)

wRAR (approved):

PR #272, opened 2026-06-23
fix _urlsplit early-exit dropping the fragment delimiter

uchiha-bug-hunter:
`_urlsplit` scans the URL once for `/ ? # [ ]` and breaks out as soon as it believes all five have been located. The exit test is a chained comparison:

    if slash_pos != question_pos != hash_pos != open_br_pos != closing_br_pos != -1:

which evaluates as `slash_pos != question_pos and question_pos != hash_pos and ...`, not "every position found". With a bracketed IPv6 netloc plus a query the chain becomes true at the `?`, before the `#` is seen:

    >>> urlsplit('http://[::1]/p?a=1#frag').fragment
    'frag'
    >>> from w3lib._url import _urlsplit
    >>> _urlsplit('http://[::1]/p?a=1#frag').fragment
    ''

So `hash_pos` stays `-1` and the fragment is folded into the query. This surfaces in the public API:

    >>> canonicalize_url('http://[::1]/p?a=1#frag')   # fragment should be dropped
    'http://[::1]/p?a=1%23frag'

Replacing the condition with a real all-found check makes the early exit match a full scan. Verified equal to a no-break full scan over 3M randomized inputs, and the existing suite still passes.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1432,6 +1432,15 @@ class TestCanonicalizeUrl:
             == "http://user:pass@www.example.com/do?a=1#frag"
         )
 
+    def test_remove_fragments_ipv6_host(self):
+        # The fragment delimiter must still be detected when the netloc is a
+        # bracketed IPv6 literal and the URL also has a query string.
+        assert canonicalize_url("http://[::1]/do?a=1#frag") == "http://[::1]/do?a=1"
+        assert (
+            canonicalize_url("http://[::1]/do?a=1#frag", keep_fragments=True)
+            == "http://[::1]/do?a=1#frag"
+        )
+
     def test_dont_convert_safe_characters(self):
         # dont convert safe characters to percent encoding representation
         assert (
--- a/w3lib/_url.py
+++ b/w3lib/_url.py
@@ -658,7 +658,13 @@ def _urlsplit(  # pylint: disable=too-many-locals,too-many-statements
             open_br_pos = idx
         elif char == "]" and closing_br_pos == -1:
             closing_br_pos = idx
-        if slash_pos != question_pos != hash_pos != open_br_pos != closing_br_pos != -1:
+        if -1 not in (
+            slash_pos,
+            question_pos,
+            hash_pos,
+            open_br_pos,
+            closing_br_pos,
+        ):
             break
 
     if url[:2] == "//":


AdrianAtZyte:
Sonnet reports the following:

---

**Top issues to address before merging:**

**Bug 1 (html.py:233) — Security-relevant regression.** The anti-backtracking fix (\[^<>\]\* instead of .\*?) was correct for preventing ReDoS, but it introduced a bypass: any tag with a literal < in an attribute value is now invisible to remove\_tags\_with\_content. Real scraped HTML does contain this. Worth documenting as a known limitation or adding a note in the commit.

**Bug 2 (url.py:346) — Easy fix.** Change the early-return to:

Plain textANTLR4BashCC#CSSCoffeeScriptCMakeDartDjangoDockerEJSErlangGitGoGraphQLGroovyHTMLJavaJavaScriptJSONJSXKotlinLaTeXLessLuaMakefileMarkdownMATLABMarkupObjective-CPerlPHPPowerShell.propertiesProtocol BuffersPythonRRubySass (Sass)Sass (Scss)SchemeSQLShellSwiftSVGTSXTypeScriptWebAssemblyYAMLXML`   return base if (not keep_fragments or not fragment) else f"{base}#{fragment}"   `

**Bug 3 (url.py:524) — One-character fix.** Remove \[:4\]:

Plain textANTLR4BashCC#CSSCoffeeScriptCMakeDartDjangoDockerEJSErlangGitGoGraphQLGroovyHTMLJavaJavaScriptJSONJSXKotlinLaTeXLessLuaMakefileMarkdownMATLABMarkupObjective-CPerlPHPPowerShell.propertiesProtocol BuffersPythonRRubySass (Sass)Sass (Scss)SchemeSQLShellSwiftSVGTSXTypeScriptWebAssemblyYAMLXML`   if scheme.lower() != b"data":   `

**Bug 4 (\_url.py:316) — \_urlsplit loop.** The loop should start at index 0 (or 2 only for the netloc chars //). The intent was to skip the // prefix, but that's wrong for non-authority URLs. A clean fix: start the loop at 0 and special-case // detection separately.

**Bug 8 (benchmarks/init.py:7) — The PyPy skip mark is a no-op** and won't protect benchmark runs on PyPy.

---

Any thoughts?

uchiha-bug-hunter:
Went through all five against the code. They hold up, but only one is in this PR's path, and even that is a separate defect from the break-condition fix here.

**Bug 4 (`_url.py`)**: real, and it's the same scan this PR touches. The loop runs `enumerate(url[2:], 2)`, so a delimiter in the first two chars after the scheme is missed when there is no `//` authority:

```
>>> _urlsplit("x?y").query
''          # stdlib urlsplit gives 'y'
```

That's independent of the chained-comparison break though. I'd rather land this PR as the minimal break-condition fix (already validated against a no-break full scan over 3M inputs, and green) and send the loop-start fix on its own with a fresh oracle run, since moving the start to 0 touches netloc handling and needs its own validation. Can fold it in here if you'd prefer a single PR.

**Bug 2 (`url_query_cleaner`)**: real. With `keep_fragments=True` and no fragment it tacks on a bare `#`:

```
>>> url_query_cleaner("http://example.com/path", ["id"], keep_fragments=True)
'http://example.com/path#'
```

The suggested guard fixes it.

**Bug 3 (`parse_data_uri`)**: real. `scheme[:4]` makes `datax:` and `database:` parse as data URIs. Dropping `[:4]` is correct.

**Bug 8 (benchmarks)**: real. `pytest.mark.skipif(...)` on its own line is a discarded expression, it never gets applied. It would need to go into `pytestmark` as a list alongside the benchmark mark.

**Bug 1 (`remove_tags_with_content`)**: that one's a deliberate tradeoff from #271. `[^<>]*` stops at a literal `<` inside an attribute value, which is malformed per spec (should be `&lt;`), and the old `.*?` is exactly what caused the backtracking. I can add a comment documenting the limitation; reworking it to catch that case without reintroducing the blowup belongs in #271.

2, 3, and 8 are unrelated files, so I'd keep them as small separate PRs rather than piling them in here. Let me know which you want first.

AdrianAtZyte (approved):

wRAR (approved):

PR #273, opened 2026-06-24
decode percent-encoded unreserved bytes in _unquote_plus

uchiha-bug-hunter:
The query parsers `_parse_qs` and `_parse_qsl` run every key and value through `_unquote_plus`, which builds its decode table from `_safe_table()`. That call takes no argument, so it falls back to the `RFC3986_UNRESERVED` default and treats `A-Za-z0-9-._~` as safe, leaving any %41-style percent escape for an unreserved byte undecoded. The sibling `_unquote` passes `safe=b""` and decodes everything, which is why only `_unquote_plus` is affected and why reserved escapes (%2F, %3D) still decode and the existing tests stayed green. The user-facing effect is that `url_query_parameter('...?x=%41%42', 'x')` returns `'%41%42'` instead of `'AB'`, and `canonicalize_url` both double-encodes (`?a=%41` becomes `?a=%2541`) and stops treating `?a=%41` and `?a=A` as the same URL, so deduplication and any value check built on these helpers can be sidestepped with percent-encoding. I found it diffing the bytes-based decoders against `urllib.parse` over random inputs, where `_unquote` matched and `_unquote_plus` diverged on every unreserved escape. Passing `b""` to `_safe_table` restores full decoding; the added cases in `TestParseQsl` cover unreserved escapes that the existing reserved-only cases missed.


uchiha-bug-hunter:
any update?

wRAR (approved):

PR #274, opened 2026-07-02
start _urlsplit delimiter scan at 0 for authority-less URLs

uchiha-bug-hunter:
`_urlsplit` scans for delimiters with `enumerate(url[2:], 2)`. That offset skips the leading `//` of an authority, but it also runs when there is no authority, so a delimiter at index 0 or 1 is never recorded:

    >>> urlsplit("a?b")
    SplitResult(scheme='', netloc='', path='a', query='b', fragment='')
    >>> from w3lib._url import _urlsplit
    >>> tuple(_urlsplit("a?b"))
    ('', '', 'a?b', '', '')          # query merged into path

For short relative URLs (`a?b`, `?x=1`, `a#f`) the `?`/`#` stays in the path. Surfaces through the public helpers:

    url_query_parameter("a?id=1", "id")   -> None   (should be "1")
    canonicalize_url("a#frag")             -> "a%23frag"  (fragment not stripped)

#272 fixed the early-exit for the IPv6 case but the scan still starts at index 2, so this is a separate path. Start the scan at 0 unless the URL begins with `//`, in which case the existing offset still skips the authority prefix.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -915,6 +915,8 @@ class TestUrl:
         assert url_query_parameter("product.html?id=", "id", keep_blank_values=1) == ""
         # only the first one is returned
         assert url_query_parameter("product.html?id=200&id=201&id=202", "id") == "200"
+        # query delimiter at index 1 of a short relative URL
+        assert url_query_parameter("a?id=200", "id") == "200"
 
     @pytest.mark.xfail
     def test_url_query_parameter_2(self):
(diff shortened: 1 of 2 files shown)


wRAR:
(IIRC it was also found by my hypothesis tests)

wRAR (approved):

AdrianAtZyte (approved):

PR #275, opened 2026-07-04
Only IDNA-encode the host in canonicalize_url, not the whole netloc

### LINKED ISSUE #222: canonicalize_url incorrectly handles port when using hostname that requires IDNA encoding

eeshsaxena:
## Summary

Fixes #222.

`canonicalize_url()` corrupts internationalized hostnames whenever a port (or userinfo) is present:

```python
>>> from w3lib.url import canonicalize_url
>>> canonicalize_url("https://тест.тест:33")
'https://xn--e1aybc.xn--33-mlc2cdc/'   # before  (port folded into the last label)
'https://xn--e1aybc.xn--e1aybc:33/'    # after   (expected)
```

## Root cause

`_safe_ParseResult()` did `netloc = _idna_str(parts.netloc)`, passing the **entire** netloc to the `"idna"` codec. That codec expects a bare domain name and splits on `.`, so for `тест.тест:33` it treats `тест:33` as one label and punycode-encodes the port into it.

## Fix

Rebuild the netloc from its parsed parts and IDNA-encode **only the host**, keeping any userinfo and port verbatim and preserving IPv6 brackets. This mirrors what `safe_url_string()` already does for the same purpose, so the two entry points now handle IDNA hosts consistently. `_idna_str` is no longer used and its import is removed.

## Tests

Added `test_canonicalize_idns_with_port` (IDN host with a port, and with userinfo + port). Full `tests/test_url.py` and the whole suite pass (361 passed, 76 xfailed unchanged), and existing IDNA / port / userinfo / IPv6 cases are unaffected.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,13 @@
 w3lib release notes
 ===================
 
+Unreleased
+----------
+
+- :func:`~w3lib.url.canonicalize_url` no longer corrupts an
+  internationalized host when a port or userinfo is present; only the host
+  is IDNA-encoded (#275).
+
 2.4.1 (2026-03-20)
 ------------------
 
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1491,6 +1491,18 @@ class TestCanonicalizeUrl:
             == "http://xn--p8j9a0d9c9a.xn--q9jyb4c/?maxResults=5&query=%E3%82%B5"
         )
 
+    def test_canonicalize_idns_with_port(self):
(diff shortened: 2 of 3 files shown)


wRAR:
Accounts that mass-submit LLM-generated PRs are unlikely to review their submissions manually and/or address PR feedback. Closing without a review. Please don't submit further PRs.

PR #279, opened 2026-07-11
strip ascii tab and newlines in canonicalize_url for bytes input

uchiha-bug-hunter:
>>> canonicalize_url("http://exa\tmple.com/")
    'http://example.com/'
    >>> canonicalize_url(b"http://exa\tmple.com/")
    'http://exa\tmple.com/'

canonicalize_url runs the WHATWG cleanup (`_strip`, which drops ASCII tab/CR/LF from anywhere in the URL) only inside `if isinstance(url, str)`, so bytes input skips it and a tab stays in the host of the canonical form. A browser strips that tab and connects to `example.com`, so the value used for dedup/allowlist comparison no longer matches the resolved host. `safe_url_string` avoids this by running `_strip(to_unicode(url, ...))` for both types; decode bytes the same way (UTF-8, as `parse_url` already does) before `_strip`.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1607,6 +1607,11 @@ class TestCanonicalizeUrl:
             == "http://www.example.com/abc/def?x=1"
         )
 
+    def test_parse_url_parse_result(self):
+        # an already parsed url is returned as is
+        parts = parse_url("http://www.example.com/path;params?x=1#frag")
+        assert parse_url(parts) is parts
+
     def test_canonicalize_url_idempotence(self):
         for url, enc in [
             ("http://www.bücher.de/résumé?q=résumé", "utf8"),
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1518,6 +1518,12 @@ class TestCanonicalizeUrl:
             canonicalize_url("http://はじめよう.みんな/?query=サ&maxResults=5")
             == "http://xn--p8j9a0d9c9a.xn--q9jyb4c/?maxResults=5&query=%E3%82%B5"
         )
+        # non-ASCII domain with an explicit port: IDNA encoding must only
+        # apply to the host, not swallow the port separator
+        assert (
+            canonicalize_url("http://www.bücher.de:8080/?q=bücher")
+            == "http://www.xn--bcher-kva.de:8080/?q=b%C3%BCcher"
+        )
 
     def test_quoted_slash_and_question_sign(self):
         assert (
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1355,17 +1355,20 @@ class TestCanonicalizeUrl:
             == "http://www.example.com/r%C3%A9sum%C3%A9?country=%D0%A0%D0%BE%D1%81%D1%81%D0%B8%D1%8F"
         )
 
-    def test_canonicalize_url_remove_ascii_tab_and_newlines(self):
+    @pytest.mark.parametrize(
+        "url",
+        [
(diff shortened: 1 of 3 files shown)


albertedwardson:
seems as a valid fix

albertedwardson:
also, I'd remove `_safe_ParseResult` entirely and just rely on `safe_url_string` in `canonicalize_url`

uchiha-bug-hunter:
Done, _safe_ParseResult is gone and canonicalize_url now goes through safe_url_string. That also made the explicit bytes decode from this PR redundant, since safe_url_string already runs _strip(to_unicode(...)) for both input types.

Two behavior changes fell out of it, both improvements as far as I can tell. IDNA now applies to the host only, so a non-ASCII domain with a port no longer gets its colon swallowed: http://www.bücher.de:8080/ used to canonicalize to www.xn--bcher-kva.de8080. I added an assertion for that in test_canonicalize_idns, verified it fails without the change. And undecodable bytes input now gets percent-encoded (safe_url_string's percentencode error handler) instead of raising UnicodeDecodeError.

On the codspeed report: the URL is now parsed twice (once inside safe_url_string, once after to split out params), which costs about 3-4% on the benchmark cases on my machine. The 11% figure was flagged as a cross-environment comparison, so it's likely inflated.

albertedwardson:
you can improve performance a bit by extracting logic of safe_url_string into some private function, calling it, and unsplitting the result at the return


this will eliminate overhead of extra unsplitting/splitting in canonicalise_url, if you call a newly introduced private function with all the making url safe logic

uchiha-bug-hunter:
Done. The body of safe_url_string now lives in a private _safe_url_split that returns the split parts, safe_url_string just unsplits them at the return, and canonicalize_url calls _safe_url_split directly. The params split moved out of _urlparse into a shared _split_params helper so canonicalize_url can separate params from the quoted path without reparsing. On the four canonicalize_url benchmark cases, timeit locally now shows this slightly ahead of master (it was ~13% behind with the double parse).

AdrianAtZyte (approved):

tests/test_url.py
```diff
@@ -1355,6 +1355,18 @@ def test_canonicalize_url_unicode_query_string_wrong_encoding(self):
             == "http://www.example.com/r%C3%A9sum%C3%A9?country=%D0%A0%D0%BE%D1%81%D1%81%D0%B8%D1%8F"
         )
 
+    def test_canonicalize_url_remove_ascii_tab_and_newlines(self):
+        # ASCII tab and newline are removed from anywhere in the URL, matching
+        # safe_url_string() and the WHATWG/urlsplit behavior. This must not
+        # depend on whether the input is str or bytes: a tab in the host of a
+        # bytes URL was previously left in place, so the canonical form kept a
+        # host a browser would connect to differently.
+        for url in (
```
wRAR: Can you please convert this to a parametrized test?

uchiha-bug-hunter: Done, converted it to a parametrized test with str/bytes param ids.

wRAR:
Can you please look at the coverage changes? It seems that passing `ParseResult` to `parse_url()` is no longer covered and `_idna_str()` can be removed.

uchiha-bug-hunter:
Done. Removed _idna_str (dead since _safe_ParseResult went away) and added a direct test for parse_url with a ParseResult input, which was previously only covered through canonicalize_url. Ran the suite with coverage locally and the missing lines now match master exactly, so the remaining gaps are the pre-existing ones (Windows-only _url2pathname paths etc.).

wRAR (approved):

PR #281, opened 2026-07-16
add userinfo regression tests for canonicalize_url with IDN hosts

uchiha-bug-hunter:
Originally a fix for _safe_ParseResult folding userinfo and port into the IDNA host label. #279 removed that code path and routes canonicalize_url through safe_url_string, which already encodes only the host, so the fix itself is no longer needed. Rebased down to the regression tests that don't exist on master: userinfo with an IDN host, and userinfo plus port with an IDN host. The plain host-with-port case was dropped since test_canonicalize_idns already covers it.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1536,6 +1536,18 @@ class TestCanonicalizeUrl:
             == "http://www.xn--bcher-kva.de:8080/?q=b%C3%BCcher"
         )
 
+    def test_canonicalize_idns_with_userinfo(self):
+        # userinfo must survive with its "@" delimiter instead of being
+        # folded into the IDNA host label
+        assert (
+            canonicalize_url("http://user@例え.jp/path")
+            == "http://user@xn--r8jz45g.jp/path"
+        )
+        assert (
+            canonicalize_url("http://user:pass@例え.jp:8080/p")
+            == "http://user:pass@xn--r8jz45g.jp:8080/p"
+        )
+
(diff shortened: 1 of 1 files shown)


wRAR:
How does this work with #279?

uchiha-bug-hunter:
They overlap at this point. #279 started as just the bytes _strip fix, but after albertedwardson's suggestion there it now removes _safe_ParseResult entirely and routes canonicalize_url through safe_url_string, which already encodes only the host. So it fixes this bug too. I ran the three cases from this PR against the #279 head and they all come out right, though #279 only asserts the port case, not the userinfo ones.

This one is the minimal standalone fix if the bigger refactor in #279 doesn't land. If #279 goes in first, I'd rebase this down to just the extra userinfo regression tests, or close it, whichever you prefer.

wRAR:
> If #279 goes in first, I'd rebase this down to just the extra userinfo regression tests, or close it, whichever you prefer.

More tests is always welcome, as I assume they don't duplicate existing ones.

uchiha-bug-hunter:
Rebased onto master now that #279 landed. The code change is gone since canonicalize_url now goes through safe_url_string; what's left is two userinfo regression tests (user@ and user:pass@ with a port), both passing against master. I dropped the plain IDN-host-with-port assertion since test_canonicalize_idns already covers that one.

wRAR (approved):

PR #285, opened 2026-07-22
treat backslash as slash for special-scheme urls in _urlsplit

uchiha-bug-hunter:
**Backslash is not treated as an authority delimiter in _urlsplit**

Under the URL living standard "\" behaves like "/" for special schemes, so a browser reads `http://evil.com\@good.com/` as host `evil.com`, but `_urlsplit` treated the backslash as an ordinary character and reported `good.com` with userinfo `evil.com\`, disagreeing with the fetched host. This converts "\" to "/" in the authority and path of special-scheme URLs (query and fragment are left as-is), which also flips the two `\`-to-`/` reference cases that were parked as xfails in `test_url.py`.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -220,6 +220,10 @@ SAFE_URL_URL_CASES = (
         f"https://{USERNAME_ENCODED}:{PASSWORD_ENCODED}@example.com",
     ),
     ("https://@\\example.com", ValueError),
+    # "\" ends the authority of a special-scheme URL, so a "\" before "@" is
+    # not a userinfo separator: the host is what precedes the "\".
+    ("https://evil.com\\@good.com/", "https://evil.com/@good.com/"),
+    ("https://good.com\\@evil.com/", "https://good.com/@evil.com/"),
     ("https://\x80:\x80@example.com", "https://%C2%80:%C2%80@example.com"),
     # Host
     ("https://example.com", "https://example.com"),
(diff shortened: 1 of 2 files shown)


tests/test_url.py
```diff
@@ -477,12 +481,8 @@ def test_safe_url_string_encoding(
     "http://192.168.0.256",  # Invalid IP address
     "http://192.168.0.0.0",  # Invalid IP address / domain name
     "https://example.com:",  # Removes the :
-    # Does not convert \ to /
-    "https://example.com\\a",
-    "https://example.com\\a\\b",
-    # Encodes \ and / after the first one in the path
+    # Encodes / after the first one in the path
     "https://example.com/a/b",
```
wRAR: Hm, what is getting encoded here?

wRAR:
```
In [2]: safe_url_string("https://example.com/a/b")
Out[2]: 'https://example.com/a/b'
```

OK, looks like a dead example (for some time?).

wRAR: Added in 4537ff58, unlikely to be broken even then, must be a typo?

wRAR: So I would remove it *unless* there are ideas of a real useful example that still fails here.

AdrianAtZyte: I thought so, but I think if it is in `KNOWN_SAFE_URL_STRING_URL_ISSUES` and was passing it must be an actual issue. Maybe that second / is not an ASCII / but an [Unicode alternative](http://xahlee.info/comp/unicode_slash.html)?

wRAR: Apparently `KNOWN_SAFE_URL_STRING_URL_ISSUES` is not used for parametrizing directly but to classify parameters from `SAFE_URL_URL_CASES`, so additional entries in the former don't affect tests.

uchiha-bug-hunter: Right, nothing gets encoded there. That string only shows up as the output of the `example.com/a\b` case, so there was no matching input in SAFE_URL_URL_CASES for it to classify: it was a dead leftover after I dropped the backslash entries. Removed it in 618348f.

wRAR:
Can you try having complete coverage of the new function?

uchiha-bug-hunter:
Added cases for the query and fragment paths in a6d2ed3. The uncovered bit was the fragment cut; the new cases hit both orderings of "?" and "#" and also pin down that a backslash inside the query or fragment is percent-encoded rather than converted. The block is at full line and branch coverage locally now.

uchiha-bug-hunter:
Added the missing cases in a6d2ed3 so both the `?` and `#` branches are exercised: query-only, fragment-only, and both `?`/`#` orderings. The new block is at full branch coverage now. The other misses codecov shows in _url.py are pre-existing code outside this patch.

wRAR (approved):

PR #293, opened 2026-07-31
skip quoted-string parameter values in http_content_type_encoding

uchiha-bug-hunter:
This fixes a regression from #291: the trailing `(?=[ \t]*(?:;|$))` lookahead
added there drops a valid charset whenever anything follows it. On current
master:

    'text/html;charset=utf-8, text/html'          -> None
    'Content-Type: text/html; charset=utf-8\r\n'  -> None  (the docstring's own form)
    'text/html; charset="utf-8", text/html'       -> None

`html_to_unicode` then falls through to the `<meta>` prescan, which the
response body controls, even though the transport layer outranks it. Since
comma-joined duplicate headers are exactly what Scrapy hands this function,
this is worth landing before the next release.

It also fixes the pre-existing quoted-string bug the lookahead didn't touch:

    >>> http_content_type_encoding('text/html; foo="bar; charset=utf-7;"; charset=utf-8')
    'utf-7'   # master; the real parameter is utf-8

Per mimesniff's "parse a MIME type", a parameter value is a token or a
quoted-string, and a quoted-string is opaque, so the `charset=` inside `foo`'s
value is not a parameter of its own; the scan has to consume the quoted value
whole. The code comments cite the mimesniff and Fetch commit snapshots, same
style as `_url.py` and `_infra.py`.

Comma is not a parameter delimiter in mimesniff; stopping a value at `,` is an
approximation of Fetch's "extract a MIME type" for comma-joined duplicate
headers. For differing essences it keeps the first charset where Fetch would
take the last valid MIME type's; a test pins
`'text/html; charset=utf-8, text/plain; charset=utf-16'` -> `'utf-8'`.

On "why not the stdlib": `email.message_from_string(...).get_param("charset")`
is not a drop-in. It parses RFC 2045 rather than WHATWG and returns
`'utf-8, text/html'` for the comma case and `'utf-8x"; charset=utf-16'` for
the orphan-quote input an existing test pins to None.

All 20 existing assertions keep their results, and the parameter walk stays
linear on adversarial input.

### DIFF
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -98,6 +98,28 @@ class TestRequestEncoding:
         assert http_content_type_encoding("text/html; mycharset=utf-8") is None
         # invalid: orphan quote in unquoted value
         assert http_content_type_encoding('text/html; charset=utf-8x"') is None
+        # the parameter is still found when the header does not end right
+        # after it
+        assert (
+            http_content_type_encoding("text/html;charset=utf-8, text/html") == "utf-8"
+        )
+        assert (
+            http_content_type_encoding('text/html; charset="utf-8", text/html')
+            == "utf-8"
+        )
+        assert (
+            http_content_type_encoding("Content-Type: text/html; charset=utf-8\r\n")
+            == "utf-8"
+        )
+        # comma-joined headers with differing essences: the first charset
+        # wins, unlike Fetch's extract-a-MIME-type where the last valid MIME
+        # type's would
+        assert (
+            http_content_type_encoding(
+                "text/html; charset=utf-8, text/plain; charset=utf-16"
+            )
+            == "utf-8"
+        )
 
     def test_html_body_declared_encoding(self):
         for fragment in self.utf8_fragments:
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -108,6 +108,22 @@ class TestRequestEncoding:
             == "utf-8"
         )
         assert http_content_type_encoding('text/html; foo="bar; charset=utf-7"') is None
+        assert (
+            http_content_type_encoding('text/html; foo="bar; charset=utf-7;"') is None
+        )
+        assert (
+            http_content_type_encoding(
+                'text/html; foo="bar; charset=utf-7;"; charset="utf-8"'
+            )
+            == "utf-8"
+        )
+        # an invalid charset value is skipped and the scan continues
+        assert (
+            http_content_type_encoding(
+                'text/html; foo="bar; charset=utf-7;"; charset=""; charset=utf-8'
+            )
+            == "utf-8"
+        )
         # the parameter is still found when the header does not end right
         # after it
         assert (
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -60,7 +60,8 @@ class TestRequestEncoding:
         extracted = http_content_type_encoding(header_value)
         assert extracted == "iso8859-4"
         assert http_content_type_encoding("something else") is None
-        # valid quoted charset values (RFC 7231 sec. 3.1.1.1)
+        # valid quoted charset values
+        # (https://mimesniff.spec.whatwg.org/#parse-a-mime-type)
         assert (
             http_content_type_encoding('Content-Type: text/html; charset="utf-8"')
             == "utf-8"
@@ -98,6 +99,15 @@ class TestRequestEncoding:
         assert http_content_type_encoding("text/html; mycharset=utf-8") is None
         # invalid: orphan quote in unquoted value
         assert http_content_type_encoding('text/html; charset=utf-8x"') is None
+        # a quoted-string is opaque: a charset written inside another
+        # parameter's value belongs to that value
+        assert (
+            http_content_type_encoding(
+                'text/html; foo="bar; charset=utf-7;"; charset=utf-8'
+            )
+            == "utf-8"
+        )
+        assert http_content_type_encoding('text/html; foo="bar; charset=utf-7"') is None
         # the parameter is still found when the header does not end right
         # after it
         assert (
--- a/w3lib/encoding.py
+++ b/w3lib/encoding.py
@@ -27,6 +27,30 @@ _HEADER_ENCODING_RE = re.compile(
     r"(?![^\s;,])",
     re.IGNORECASE,
 )
+# https://mimesniff.spec.whatwg.org/commit-snapshots/39aa53511b13953d84fef8d4131d6f61d0ccbde6/#parse-a-mime-type
+# Parameters are ";"-separated name=value pairs whose value is either a token
+# or a quoted-string, and a quoted-string is opaque
+# (https://fetch.spec.whatwg.org/commit-snapshots/586cd2a44c2a865b37c166dc0740f3fb8bb220d6/#collect-an-http-quoted-string),
+# so it has to be consumed as a whole: a "charset=" written inside one belongs
+# to that value and is not a parameter of its own.
+_HEADER_PARAMETER_RE = re.compile(
+    r"(?:^|;)[ \t]*(?P<name>[^\s;=]+)="
+    r'(?:"(?P<quoted>[^"\\]*(?:\\.[^"\\]*)*)"(?![^\s;,])'
+    r"|(?P<token>[^;,\s]*))"
+)
+_ENCODING_LABEL_RE = re.compile(r"[\w-]+")
+
+
+def _quoted_aware_charset(content_type: str) -> str | None:
+    for match in _HEADER_PARAMETER_RE.finditer(content_type):
+        if match.group("name").lower() != "charset":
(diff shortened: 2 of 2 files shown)


AdrianAtZyte:
<details>
<summary>Opus feedback</summary>

```
Verdict

The behaviour change is right, the citation is wrong, and there's one case it silently regresses.

On the RFC reference — your instinct is correct

RFC 9110 §5.6.6 and WHATWG mimesniff §4.4 ("parse a MIME type") agree here, so the fix is sound either way. But w3lib/encoding.py:115 already points at WHATWG, and _url.py / _infra.py cite WHATWG with commit snapshots. The comment should say mimesniff §4.4 + Fetch's "collect an HTTP quoted string", pinned to a commit snapshot like _url.py:27 does. (tests/test_encoding.py:63 cites RFC 7231 from #291 — same problem, worth fixing in the same pass.)

It fixes a real regression from #291 (merged 3 days ago)

The (?=[ \t]*(?:;|$)) lookahead added last week broke more than the PR body claims. On current master:

'text/html;charset=utf-8, text/html'            -> None
'Content-Type: text/html; charset=utf-8\r\n'    -> None   # the docstring's own form
'text/html; charset="utf-8", text/html'         -> None

Falling through to the <meta> prescan when the transport layer said utf-8 is a genuine bug, and comma-joined duplicate headers are exactly what Scrapy hands this function. This should land before the next release, and the description should call it out as a regression fix.

The one thing I'd push back on: comma handling

Comma isn't a delimiter in mimesniff at all. Comma-joined values are Fetch's extract a MIME type, which says the last valid MIME type wins, and charset carries forward only within a run of the same essence. The PR bakes , into the terminator set and returns the first charset:

'text/html; charset=utf-8, text/plain; charset=utf-16'
    master -> 'utf-16'   (matches Fetch, by accident)
    PR     -> 'utf-8'    (diverges)

So it fixes the same-essence comma case and regresses the differing-essence one. A single-string API can't implement extract-a-MIME-type properly (no essence tracking), and I would not ask for that — but the comment should say the comma handling approximates Fetch's extract-a-MIME-type rather than implying it's parameter syntax, and there should be a test pinning whichever answer you pick for that input.

Divergences I'd accept as-is

All stricter than spec, so they fall back to the <meta> prescan rather than trusting something wrong:

- charset="utf-8 (unterminated) → spec utf-8, PR None
- charset="utf\-8" (no unescaping) → spec utf-8, PR None
- charset="utf-8"x (trailing garbage) → spec utf-8, PR None

One goes the other way: mimesniff is first-wins by parameter name, but the PR skips an invalid charset and keeps scanning, so charset=""; charset=utf-8 → utf-8 where the spec says nothing. Marginal; a comment noting it's deliberate would do.

Why not stdlib

The PR body's own email.get_param demo invites "just use the stdlib". I checked — it isn't a drop-in: 'utf-8, text/html' for the comma case, 'utf-8x"; charset=utf-16' for the orphan-quote case (which an existing test pins to None), and it's RFC 2045 rather than WHATWG. The hand-rolled walk is justified; saying so in the PR would pre-empt the question.

Regex safety is fine, by the way — the quoted group is the unrolled-loop form, and the failing-lookahead path can't backtrack into it, so the linear claim holds. Existing 19 tests plus the 6 new assertions pass.
```

</details>

uchiha-bug-hunter:
All fair, addressed in 50f5f03. The comments now cite the mimesniff and Fetch commit snapshots like _url.py does, and the RFC 7231 line in test_encoding.py from #291 got the same treatment. The comma handling is now documented as an approximation of Fetch's extract-a-MIME-type, with a test pinning 'text/html; charset=utf-8, text/plain; charset=utf-16' -> 'utf-8', and the skip-invalid-charset-and-continue behavior is marked deliberate. Also reworded the description as a regression fix from #291 and added the why-not-stdlib note there.

AdrianAtZyte:
<details>
<summary>Opus feedback</summary>

```
Verdict

The behaviour is right and the regression fix should land before the next release. But the PR bundles two fixes with very different cost/benefit, and the 15% is entirely attributable to the cheaper-value one. It's avoidable.

The two fixes

(a) The #291 regression — real and common. Verified on master:

'text/html;charset=utf-8, text/html'          -> None
'Content-Type: text/html; charset=utf-8\r\n'  -> None   # the docstring's own form
'text/html; charset="utf-8", text/html'       -> None
'text/html; charset=utf-8, text/plain; ...'   -> 'utf-16'

Root cause is one token: the (?=[ \t]*(?:;|$)) lookahead. It exists only to reject trailing garbage (charset=utf-8x"), and $ can't see past \r. Replacing it with (?![^\s;,]) — "next char is whitespace, ;, , or end" — fixes all four at zero cost.

(b) Quoted-string opacity — foo="bar; charset=utf-7;"; charset=utf-8. Correct per mimesniff, but it needs a server emitting a quoted parameter value that itself contains charset=. This is the fix that costs the 15%, because it forces the targeted charset= search to become a full left-to-right walk of every parameter.

Yes, you can have both for ~0%

A match inside a quoted-string must be preceded by that string's opening quote. So if no " precedes the match, the fast answer is provably correct; only otherwise do you pay for the walk:

_HEADER_ENCODING_RE = re.compile(
    r"(?:^|;)[ \t]*charset="
    r'(?:"([\w-]+)"|([\w-]+))'
    r"(?![^\s;,])",
    re.IGNORECASE,
)
...
    match = _HEADER_ENCODING_RE.search(content_type)
    if match:
        if content_type.find('"', 0, match.start()) < 0:
            return resolve_encoding(match.group(1) or match.group(2))
        return _quoted_aware_charset(content_type)   # the PR's walk, unchanged

Measured on the two codspeed cases, isolated from resolve_encoding, best-of-9 interleaved:

┌────────┬─────────┬───────────┐
│        │ µs/iter │ vs master │
├────────┼─────────┼───────────┤
│ master │ 2.36    │ —         │
├────────┼─────────┼───────────┤
│ PR 293 │ 3.66    │ +55%      │
├────────┼─────────┼───────────┤
│ above  │ 2.63    │ +11%      │
└────────┴─────────┴───────────┘

(CodSpeed's 17% is diluted by resolve_encoding; on the parsing itself the PR is much worse.) The residual 11% is a single str.find on the match path — headers with no charset don't pay it at all, since a failed targeted search implies a failed walk.

Verified: PR's 6 new assertions plus all 19 existing ones pass, full suite 408 passed, and 200k fuzzed inputs over ; , " = \ \t charset utf-8 \r\n produce zero divergences from the PR's walk.

The "deliberately laxer" comment

Worth pushing back on — the behaviour is neither deliberate nor consistent:

'text/html; charset=""; charset=utf-8'     -> 'utf-8'   # skipped, scan continued
'text/html; charset=bogus; charset=utf-8'  -> None      # not skipped

The skip only happens for values that fail the label regex; a well-formed but unresolvable label still returns None immediately. And master already behaves this way — it's an artifact of the regex not matching, now retroactively labelled as intent. Either make it consistent (continue when resolve_encoding returns None) or drop the claim of deliberateness and just note it as inherited. I'd lean toward dropping the comment; nobody is relying on either half.

Minor

- Dropping re.IGNORECASE in favour of match.group("name").lower() allocates a string per parameter on the walk path. Keep the flag and compare against a literal, or match [Cc][Hh]... — irrelevant if the walk becomes the rare path.
- The comma approximation and its test are fine as landed.

Suggested shape

Split it: land (a) as a standalone regression fix — three-line diff, tests for the comma/CRLF/quoted-comma cases, no perf delta, cherry-pickable into a patch release. Then (b) on top with the fast-path guard, so the walk is reviewed on its own merits rather than as the price of unbreaking master.
```

</details>

uchiha-bug-hunter:
Done, split as suggested and rebased on master. The first commit is just the lookahead swap to (?![^\s;,]) plus tests for the comma/CRLF/quoted-comma cases and the differing-essence pin, so it's cherry-pickable on its own. The second adds the walk behind your guard, unchanged except one detail: the quoted branch's terminator is now the same (?![^\s;,]) as the fast regex, because the old [ \t]*(?:[;,]|$) lookahead disagreed with the fast path on a quoted value followed by \r\n. With that aligned, 200k fuzzed inputs over the same alphabet give zero divergences between the two paths for me as well. Dropped the deliberately-laxer comment, you're right that it was a regex artifact, behavior unchanged. Left the .lower() in the walk since it's the rare path now, and the codspeed inputs all take the fast path, so the benchmark delta should be gone.

AdrianAtZyte:
Could you complete coverage?

uchiha-bug-hunter:
Done in 296d3e2. The uncovered lines were all in the slow walk, which the existing tests only exercised through unquoted charset values: a quoted charset resolved by the walk, an invalid value making the scan continue, and the walk finding no charset at all. encoding.py is at 100% line and branch coverage locally now.

AdrianAtZyte (approved):

wRAR (approved):

PR #301, opened 2026-08-09
remove ascii tab and newline from the url in _urlsplit

uchiha-bug-hunter:
parse_url("http://exa\tmple.com/p").hostname  ->  'exa\tmple.com'
    urlparse("http://exa\tmple.com/p").hostname   ->  'example.com'

`_urlsplit` reimplements `urllib.parse.urlsplit` but skips its "remove all ASCII tab or newline" step, only lstripping leading C0/space. A tab, CR or LF left in the authority survives, so `parse_url` (also `url_query_parameter`, `add_or_replace_parameter`) reports a host a browser never connects to; a tabbed port like `http://example.com:8\t0/` raised ValueError where `urlparse` returns port 80. #279 added the strip to `canonicalize_url`'s bytes path only, so `parse_url` still diverged from `urlparse`. Drop tab/CR/LF from the url and scheme at the top of `_urlsplit`, matching `urlsplit`; `safe_url_string`/`canonicalize_url` already `_strip` first and are unchanged.

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1661,6 +1661,24 @@ class TestCanonicalizeUrl:
         assert parse_url(url).path == urlparse(url).path
         assert parse_url(url).params == urlparse(url).params
 
+    @pytest.mark.parametrize(
+        "url",
+        [
+            "http://exa\tmple.com/p",
+            "http://exa\nmple.com/p",
+            "http://exa\rmple.com/p",
+            "https://good.com\t.evil.com/a\r\nb?q=a\tb#f\tr",
+            "http://example.com:8\t0/",
+        ],
+    )
+    def test_parse_url_remove_ascii_tab_and_newlines(self, url):
+        # urlsplit drops every ASCII tab and newline from the URL, so a tab in
+        # the host cannot survive into parse_url()'s netloc and report a host a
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
Any chance we can get a fix without the performance penalty?

uchiha-bug-hunter:
Sure. The cost was the dict-based translate running unconditionally on every _urlsplit call, even for clean URLs, which is all the benchmarks parse. Pushed a guard that checks for `\t`/`\n`/`\r` first and only translates on a hit, same pattern `_strip` already uses. Locally the cold benchmark delta vs master goes from +10-14% to within noise; the membership checks cost about 80ns per URL vs about 1.1us for the translate.

AdrianAtZyte:
Nice! What about coverage?

uchiha-bug-hunter:
The uncovered lines were the scheme-argument branch: nothing in the public API passes a default scheme, so a scheme containing a tab was never exercised. Added a direct _urlsplit test for it (test_url.py already imports other private helpers), and branch coverage now shows all four patch lines hit locally.

AdrianAtZyte (approved):

wRAR (approved):

PR #303, opened 2026-08-11
strip script and noscript content in get_base_url

uchiha-bug-hunter:
>>> get_base_url("<script>var t = \"<base href='http://evil.example/'>\";</script>", "https://good.example/")
    'http://evil.example/'

get_base_url only runs remove_comments before scanning for `<base href>`, so a `<base>` written as text inside `<script>` or `<noscript>` is picked up and used to resolve relative URLs. A browser never parses tags inside those elements, so it ignores such a base and resolves against the real one; a consumer that reflects untrusted markup into a script/noscript block ends up resolving links against an attacker-chosen base.

get_meta_refresh already drops script/noscript for the same reason. This does the same in get_base_url before the search. A `<base>` that follows the ignored content is still returned.

### DIFF
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -474,6 +474,33 @@ class TestGetBaseUrl:
             )
             == "http://example.org/found/"
         )
+        # an unterminated <script> swallows the rest of the document, as in a
+        # browser
+        assert (
+            get_base_url(
+                """<script><base href="http://evil.example/">""",
+                baseurl,
+            )
+            == "https://example.org"
+        )
+        assert (
+            get_base_url(
+                """<noscript foo="bar"><base href="http://evil.example/">""",
+                baseurl,
+            )
+            == "https://example.org"
+        )
+
+    def test_base_url_split_by_comment(self):
+        # A comment inside the <base> tag is not stripped: browsers do not
+        # parse comments within a tag, so the tag is broken and ignored.
+        assert (
+            get_base_url(
+                """<base h<!--c-->ref="http://example.com/">""",
+                "https://example.org",
+            )
+            == "https://example.org"
+        )
 
     def test_relative_url_with_absolute_path(self):
         baseurl = "https://example.org"
--- a/tests/test_html.py
+++ b/tests/test_html.py
@@ -447,6 +447,34 @@ class TestGetBaseUrl:
             == "http://example_3.com/"
         )
 
+    def test_base_url_in_script(self):
+        baseurl = "https://example.org"
+        # A browser does not parse tags inside <script>/<noscript>, so a <base>
+        # written there is ignored and relative URLs resolve against baseurl.
+        assert (
+            get_base_url(
+                """<script>var t = "<base href='http://evil.example/'>";</script>""",
+                baseurl,
+            )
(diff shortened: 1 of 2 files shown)


AdrianAtZyte:
> Two things.
>
> The strip is incomplete: `remove_tags_with_content` needs a closing tag, so an unterminated `<script>` still gets through, and a browser swallows the rest of the document there.
>
> ```pycon
> >>> get_base_url('<script><base href="http://evil.example/">', "https://good.example/")
> 'http://evil.example/'
> ```
>
> This is the same hole that #298 fixed for unterminated `<meta>` tags. I think `get_meta_refresh` has it too, since it strips the same way, but that is a separate issue.
>
> The performance hit is avoidable. Both `remove_comments` and `remove_tags_with_content` rewrite the whole document, and we only need the first `<base href>`. A single scan that stops at the first hit does both jobs and comes out ahead of master:
>
> ```python
> _base_scan_re = re.compile(
>     r"""
>       |$)
>     | <(?P<t>script|noscript)\b[^<>]*>.*?(?:</(?P=t)>|$)
>     | <base\s[^<>]*href\s*=\s*["']\s*(?P<url>[^"'\s]+)\s*["']
>     """,
>     re.IGNORECASE | re.DOTALL | re.VERBOSE,
> )
>
>
> def get_base_url(text, baseurl="", encoding="utf-8"):
>     utext = to_unicode(text, encoding)
>     for m in _base_scan_re.finditer(utext):
>         if url := m.group("url"):
>             return urljoin(
>                 safe_url_string(baseurl), safe_url_string(url, encoding=encoding)
>             )
>     return safe_url_string(baseurl)
> ```
>
> Ignorable regions come first in the alternation, so a `<base>` inside one is consumed before it can match. Over the five benchmark cases on my machine:
>
> ```
> master     113.21 us/loop
> this PR    148.23 us/loop   (+31%)
> combined   105.36 us/loop   (-7%)
> ```
>
> The existing tests pass with it, along with your three new assertions and the unterminated `<script>` case above, and it stays linear on `"<script " * 50000`, `"<!--" * 50000` and the existing `"<base " * 30000` case.
>
> One behavior change worth calling out: master resolves `<base h<!--c-->ref="http://example.com/">` because it strips comments before matching, and this does not. Browsers do not strip comments inside a tag either, so I think that is a fix, but please add a test for it either way.

uchiha-bug-hunter:
Good points, both confirmed here. Switched to your combined scan and added tests for the unterminated <script> and <noscript> cases.

Also added the test for the comment-split tag: `<base h<!--c-->ref=...>` now resolves to the fallback baseurl, and I agree that's a fix, since a browser tokenizes that as a broken tag and never sees an href.

On the five benchmark cases this comes out at parity with master on my machine (17.2 vs 16.7 us/loop, vs +31% for the old approach), and the `"<script " * 50000`, `"<!--" * 50000` and `"<base " * 30000` inputs all stay linear.

You're right that get_meta_refresh has the same unterminated-tag hole via remove_tags_with_content; I'll look at that in a separate PR.

AdrianAtZyte (approved):

wRAR (approved):

PR #313, opened 2026-08-31
Some new tests and cleanup/fixes.

wRAR:


tests/benchmarks/test_benchmark_encoding.py
```diff
@@ -20,6 +21,11 @@
 
     from tests.benchmarks import CasesMapType
 
+pytestmark = [
+    pytest.mark.benchmark,
+    pytest.mark.skipif("PyPy" in sys.version, reason="CodSpeed doesn't support PyPy"),
+]
```
AdrianAtZyte, line 27: 💄 Could we define the list in `__init__.py` and then set `pytestmark = <list var>` here instead?

pyproject.toml
```diff
@@ -88,6 +88,7 @@ enable = [
     "useless-suppression",
 ]
 disable = [
+    "duplicate-code",
```
AdrianAtZyte, line 91: Would it make sense to disable it only for tests/?

AdrianAtZyte (approved):

PR #320, opened 2026-09-04
resolve percent-encoded dot segments in safe_download_url

uchiha-bug-hunter:
>>> safe_download_url("http://www.example.org/dir/%2e%2e/secret")
    'http://www.example.org/dir/%2e%2e/secret'

`%2e`, `.%2e`, `%2e.` and `%2e%2e` are single/double-dot path segments under the URL living standard (ASCII case-insensitive), so a compliant client resolves that URL to `/secret`, outside `/dir/`: the traversal the function is documented to neutralize survives in encoded form, while its literal spelling `/dir/../secret` is stripped. `canonicalize_url` already resolves these because it percent-decodes the path before `normpath`. Here the encoded dot segments are decoded segment-wise before `normpath`, behind a `"%" in path` test so percent-free paths keep the current path untouched; segments that merely contain `%2e` (e.g. `a%2eb`) are left alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XHtkcqVANsVciDasEK5Ys1

### DIFF
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -980,6 +980,30 @@ class TestUrl:
             == "http://www.example.org/%A3?%C2%A3"
         )
 
+    def test_safe_download_url_encoded_dot_segments(self):
+        # "%2e", ".%2e", "%2e." and "%2e%2e" are the percent-encoded forms of
+        # the single-dot and double-dot path segments of the URL living
+        # standard, resolved by clients like "." and "..".
+        assert (
+            safe_download_url("http://www.example.org/dir/%2e%2e/secret")
+            == "http://www.example.org/secret"
+        )
+        assert (
+            safe_download_url(
+                "http://www.example.org/%2E%2E/%2E%2E/images/%2e%2e/image"
+            )
+            == "http://www.example.org/image"
+        )
+        assert (
+            safe_download_url("http://www.example.org/dir/.%2e/%2e./a/%2e/b")
+            == "http://www.example.org/a/b"
+        )
+        # Segments that merely contain "%2e" are not dot segments.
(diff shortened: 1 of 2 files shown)


AdrianAtZyte (approved):
Looks good to me. Although I wonder if safe_download_url() is really the ideal target for this change, or if it should be applied to some underlying function.

uchiha-bug-hunter:
Had a look at where else it could go. Only two functions resolve dot segments at all, canonicalize_url and safe_download_url, and each does its own `_parent_dirs.sub("", posixpath.normpath(path))`. canonicalize_url already handles the encoded forms for free because it percent-decodes the whole path before normpath (`canonicalize_url("http://example.org/dir/%2e%2e/secret")` already gives `http://example.org/secret` on master). safe_download_url can't decode like that since it has to keep the path percent-encoded, so it's the one place where a literal `..` is resolved but `%2e%2e` isn't.

Pushing it further down into safe_url_string / _safe_url_split doesn't seem right: that function doesn't resolve literal dot segments either (`safe_url_string("http://example.org/dir/../secret")` comes back unchanged) and is documented to return an equivalent URL, so rewriting `%2e%2e` to `..` there would change the URL without resolving it, for every caller.

If you'd prefer one shared spot, I can pull the normpath + `_parent_dirs` step out of both functions into a `_resolve_dot_segments(path)` helper that also knows the encoded forms, either here or as a follow-up.

Also rebased on master to clear the conflict with #311; the decode step now sits in front of the normalized_path block from that change, no other differences.

wRAR (approved):

PR #326, opened 2026-09-13
Fix empty fragment handling in url_query_cleaner(), fix the parameterlist hint

wRAR:


w3lib/url.py
```diff
@@ -334,7 +334,7 @@ def url_query_parameter(
 
 def url_query_cleaner(
     url: str | bytes,
-    parameterlist: str | bytes | Sequence[str | bytes] = (),
```
wRAR, line 337: Bytes here were never converted and thus never matched anything, adding the conversion makes the benchmark 17% worse.

AdrianAtZyte (approved):

PR #328, opened 2026-09-14
Remove unused "safe" from _unquote().

wRAR:
@albertedwardson I'd want your input here.

The `safe` arg of `_unquote()` was only used in `_url2pathname()` where, AFAIK, it doesn't make sense as paths shouldn't have percent encoding, so I removed it in #327. The arg of `_unquote_plus()` wasn't used at all. So here I'm removing the related code completely, and it's a non-trivial amount so I want you to say if this isn't really needed. Similarly I removed the `safe=b"/"` arg from the test, as it makes the test fail and the stdlib `unqoute()` doesn't have this arg.


wRAR:
(hypothesis test failures happen on master too with this input but I'll look into them)

AdrianAtZyte (approved):

albertedwardson:
Hm, filesystem paths themselves generally **shouldn't** contain URI percent-encoding, but they and their `file URI` representation absolutely **can**

So I don't think it was OK to completely remove passing safe chars to `_unquote` in `_url2pathname()` in #327 in the first place...

Decoding `%2F` changes the path structure. For example:

```python
>>> file_uri_to_path("file://localhost/some/dir/my%2Ffile%40.txt")
'/some/dir/my%2Ffile%40.txt'
```

If we blindly unquote everything, this becomes:

```python
>>> file_uri_to_path("file://localhost/some/dir/my%2Ffile%40.txt")
'/some/dir/my/file@.txt'
```

which is a different filesystem path.

After digging into the URI/path semantics, though, I think `_PATH_SAFE_CHARS` is too broad for this purpose. We should still decode ordinary percent-encoded characters, e.g. `%41` → `A`, and only preserve encoded path delimiters

So I'd suggest keeping the `safe` support in `_unquote` and passing`safe=b"/"` in `_url2pathname()` as the unquoting preservation set rather than `_PATH_SAFE_CHARS`

wRAR:
> Hm, filesystem paths themselves generally shouldn't contain URI percent-encoding, but they and their file URI representation absolutely can

Right. And #327 doesn't remove it, of course, I just wrote a wrong thing :facepalm: 

> file_uri_to_path("file://localhost/some/dir/my%2Ffile%40.txt")
> '/some/dir/my/file@.txt'

This indeed looks wrong. But it's the behavior from 2.4.1, in other words, it's the behavior where `urllib.request.url2pathname()` and `w3lib._url._url2pathname()` differ (before #327).

```python
In [1]: from urllib.request import url2pathname

In [2]: url2pathname('file://localhost/some/dir/my%2Ffile%40.txt')
Out[2]: 'file://localhost/some/dir/my/file@.txt'

In [3]: url2pathname('file:///localhost/some/dir/my%2Ffile%40.txt')
Out[3]: 'file:///localhost/some/dir/my/file@.txt'

In [4]: url2pathname('/some/dir/my%2Ffile%40.txt')
Out[4]: '/some/dir/my/file@.txt'
```

`urllib.request.url2pathname()` is documented to use `urllib.parse.unquote()` which doesn't have an option to provide safe chars, while our reimplementation (before this PR) does.

So I guess there are two questions: what is the "correct" way to unquote the path segment of an URL and whether we want to diverge from the stdlib on the related cases.

For the second, I know that the `urllib.parse` docstring refers to the Hyrum's Law when listing the standards it doesn't fully comply to, so I think it's fine for us to either comply to the standards only or to match stdlib, whatever we prefer.

For the first, the URL living standard doesn't seem to talk much about unquoting paths. A parsed path is a list of components with no percent-decoding (`file:///ada/Analytical%20Engine/README.md` is parsed as `["ada", "Analytical%20Engine", "README.md"]` in the examples), split on a literal slash char, and decoding the components, I assume, is left to the caller. But we don't work with components but with the reconstituted path, and it makes some sense to unquote everything there instead of everything but `%2F`.

Too bad that the only thing Google can find about `url2pathname` and `2f` is https://github.com/nltk/nltk/security/advisories/GHSA-p4gq-832x-fm9v (I haven't read it).

Thoughts?


Edit: funnily, the URL living standard does have a special handling for path components being fully or partially percent-encoded "." and ".." (#320 was about that).

AdrianAtZyte:
After some discussion with an LLM on the topic, I am worse off than I was before, and with a headache.

> 1. **Is #328 OK — can `safe` go?** Yes, merge as is. Nothing passes it; the one caller that needs a restriction (`_unquotepath()`, for `canonicalize_url()`) implements it by pre-escaping `%25`/`%2F`/`%3B`, not via `safe`. Keeping a parameter in two hot decode loops for an option nobody selects is the worst outcome, and albertedwardson's `safe=b"/"` would hand callers a path that cannot exist on any filesystem.
>
> 2. **What is the "correct" way to unquote a URL path?** Split into segments first, then decode each — RFC 3986 §2.4, because otherwise "the data may be mistaken for component delimiters". For `file:` that leaves a decoded `/` inside a segment, which is unrepresentable, so the faithful options are reject or substitute; re-joining is the one thing the RFCs rule out. No spec defines the URI→path mapping: RFC 8089 punts to implementations (§2, Appendix D) and only warns (§5), and the WHATWG living standard never mentions a filesystem — it does make `%2e%2e` a dot segment but never makes `%2F` a delimiter.
>
> 3. **Diverge from the stdlib?** Not in `_url2pathname()` — stdlib parity is its stated purpose, and Firefox and rust-url decode like CPython does. Put the divergence in `file_uri_to_path()` instead: raise `ValueError` on `%2F` (and `%5C` on Windows), like Node. Basis: every `%2F` we see is foreign input, since `path_to_file_uri()` emits `%252F`, so round-trips are unaffected; and Scrapy feeds `request.url` straight into it, where `safe_url_string()` and `canonicalize_url()` both preserve `%2F` by design — `file:///allowed/x%2F..%2F..%2Fetc/passwd` survives both and lands as `/etc/passwd`.

It feels, however, like there is no clear path here. I am surprised not even Firefox and Chromium behave the same in something as fundamental as this :expressionless: 

```
┌───────────────────────────────────┬──────────────────────────────────────────────────────────────────────┐
│                                   │                           file:///a/b%2Fc                            │
├───────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│ CPython url2pathname() 3.9–3.14,  │ /a/b/c                                                               │
│ Path.from_uri()                   │                                                                      │
├───────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│ rust-url Url::to_file_path()      │ /a/b/c — splits into segments, decodes each, then joins with the     │
│                                   │ separator (both the POSIX and Windows branches)                      │
├───────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│ Firefox 152                       │ opens /a/b/c                                                         │
├───────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│ Chrome 153                        │ not found                                                            │
├───────────────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│ Node 25 fileURLToPath()           │ ERR_INVALID_FILE_URL_PATH                                            │
└───────────────────────────────────┴──────────────────────────────────────────────────────────────────────┘
```

wRAR:
> For file: that leaves a decoded / inside a segment, which is unrepresentable, so the faithful options are reject or substitute

Well I like this perspective :)

albertedwardson:
After some research, I'm fine with dropping this and keeping `_unquote()` without `safe`. I don't see any other use cases for `safe` in `_unquote()` either 

My concern about `%2F` was mainly that decoding it as `/` can change the path structure. But since there are no clear guarantees around `file:` URL to filesystem path mapping - I don't think we need to solve that here.

wRAR:
Thanks @albertedwardson

wRAR:
As the actual code changes were subsumed into #344, closing this.

albertedwardson:
For context, here is what the WHATWG says regarding file URLs:

WHATWG spec for fetch: https://fetch.spec.whatwg.org/#scheme-fetch

> 7. "file"
For now, unfortunate as it is, file: URLs are left as an exercise for the reader.
> 
> When in doubt, return a network error.

WHATWG matrix chat: https://matrix.to/#/!AGetWbsMpFPdSgUrbs%3Amatrix.org/%24wtuLWgQ5NKDGgGDqw6EUvJxr551S3G5-Kiv-jPeEyXI

> This is currently left to the implementation as it depends on the platform. We have discussed giving more detail about this in the Fetch specification (which partially defines how to go from a URL to a resource), but nobody has done the work.

PR #336, opened 2026-09-17
Add a max_scan parameter to get_base_url() and get_meta_refresh()

AdrianAtZyte:
Based on the measurements I gathered locally so far about https://github.com/scrapy/scrapy/pull/8178, I think supporting the limit may be worth it, but I think we should make it a suggestive parameter (to be exposed as a Scrapy setting), so that users can tune it in Scrapy for performance vs correctness, and so that in w3lib we can ignore it if e.g. at any point in the future we get a better implementation that requires the entire body.


w3lib/html.py
```diff
@@ -405,17 +405,46 @@ def unquote_markup(
...
+    """Return the position in *text* where a scan limited to *max_scan*
+    characters may stop, extended past a tag that starts before that limit and
+    ends after it."""
+    if max_scan is None:
+        return len(text)
+    if max_scan < 0:
+        raise ValueError(f"max_scan must not be negative, got {max_scan!r}")
+    if max_scan >= len(text):
+        return len(text)
+    if text.rfind("<", 0, max_scan) > text.rfind(">", 0, max_scan):
```
wRAR: Are we fine with these matching stuff in comments and attr values?

wRAR (approved):

PR #343, opened 2026-09-21
Do not end a tag at an angle bracket inside a quoted attribute value

AdrianAtZyte:


wRAR:
I see this and #336 seem change the same tests, are they alternative fixes?

AdrianAtZyte:
They indeed overlapped. #336 had its own `_TAG_BODY`, an earlier take where any quote opens a value, while here a quote only opens one right after an `=`. I have moved the base and meta part into this PR, to keep things the same. #336 now merges this branch and is only about `max_scan`.

wRAR (approved):

